PR #8, opened 2019-08-22
Use go mod & update dependencies

crandles:
`go mod` vendors a little differently, hence the deletions. While I was looking at the dependencies, I went ahead and updated to the latest (I could roll that back if desired).


dougbtv:
@crandles -- this is awesome! Thanks a bunch for the go mod update... hugely nice one. Thanks for the contribution.

PR #23, opened 2019-11-26
Changes to use CR backend by default

dougbtv:
* Updates daemonset to include `IPPool` CRD
  * Also removes CRD definition from its own file.
* Updates daemonset to use `:latest` tagged docker image
* README updates to reflect CR backend as default
* Updates daemonset to use Kubernetes v1.16+ apiVersion for daemonset
* Updates README to reflect Kubernetes v1.16+ compatibility for daemonset


crandles:
The CRD file is still being autogenerated by one of the make steps, I think that is causing the duplicate error.

We could:
- disable the yaml generation for now
- move these artifacts to a directory (easy to apply)
- helm chart (I have one in use internally)
- bash to append output of files / rm extra file

dougbtv:
@crandles thanks for the pointer! ...awesome, what I did for now was separate out to where we were, and just update the docs to do a `kubectl create -f ... -f ...` with multiple `-f`'s for now.

The helm install should also be pretty sweet. At some point I'm thinking I'll put together an operator to lifecycle manage it, but, the helm install is probably... more appropriate at the moment. At some point if you're down to mix up the helm charts that you have for an upstream style install, they're gladly appreciated, but... no rush there at all, and this honestly isn't too bad (still just a couple line copy/paste)

dougbtv:
Just a heads up that I removed a line from .travis.yml that should probably go back, it's a git diff after code generation, in theory... this shouldn't have required a regeneration in this change. I filed an issue to track it #24

PR #29, opened 2020-01-03
Adds file-based 'global' configuration for NetworkAttachmentDefinition use

dougbtv:
Attempts to simplify what's needed in the CNI `ipam` section when using NetworkAttachmentDefinitions, as when you're using them... You might be creating many CNI configurations, and therefore don't want to have to repeat all of the same information in the `ipam` section for whereabouts many times (e.g. specifying kubernetes mode and kubeconfig every time, or logging options, or etcd options, etc etc).

More detail on the usage in the README.


pkg/config/config.go
```diff
@@ -42,6 +47,45 @@ func LoadIPAMConfig(bytes []byte, envArgs string) (*types.IPAMConfig, string, er
...
+
+			jsonFile, err := os.Open(confpath)
+
+			if err != nil {
+				return nil, "", fmt.Errorf("Error opening flat configuration file @ %s with: %s", confpath, err)
+			}
+
+			defer jsonFile.Close()
+
+			jsonBytes, _ := ioutil.ReadAll(jsonFile)
```
crandles: Uncaught error; I'd return it.

dougbtv: Thanks for the eye, good call, got it in the most recent commit

PR #41, opened 2020-05-31
Fixed to drop broadcast and first network address when the range is cidr

aneeshkp:
Fixed For whereabouts DHCP testing, if setup range: 192.168.21.100/30 in yaml file,  then 192.168.22.100/30 is network address, 192.168.22.101/30 to 192.168.22.102/30 are host addresses, 192.168.22.103/30 is broadcast address, the pod should only get 192.168.22.101 and 192.168.22.102, but right now one pod can get network address 192.168.22.100

### DIFF
--- a/pkg/allocate/allocate_test.go
+++ b/pkg/allocate/allocate_test.go
@@ -2,6 +2,7 @@ package allocate
 
 import (
 	"fmt"
+	"math/big"
 	"net"
 	"testing"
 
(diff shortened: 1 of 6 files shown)


pkg/allocate/allocate_test.go
```diff
@@ -15,34 +16,117 @@ func TestAllocate(t *testing.T) {
 }
 
 var _ = Describe("Allocation operations", func() {
-	It("creates an IPv4 range properly", func() {
+	It("creates an IPv4 range properly for 30 bits network address", func() {
```
dougbtv, line 19: Thank you for the thorough test additions!

PR #45, opened 2020-07-10
Make ds use hostNetwork by default

ashish-billore:
This patch makes whereabouts pods deployed as ds to use
hostNetwork by default. This is helpful as CNI and core
infra components use hostnetwork and hence avoid issues
such as fault in CNI config etc and also making debugging
much easier.

### DIFF
--- a/doc/daemonset-install.yaml
+++ b/doc/daemonset-install.yaml
@@ -57,6 +57,7 @@ spec:
         app: whereabouts
         name: whereabouts
     spec:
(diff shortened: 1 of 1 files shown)


ashish-billore:
@crandles @dougbtv @nicklesimba @aneeshkp Please suggest your opinion about this patch, thanks for the help!

ashish-billore:
Thanks @crandles @dougbtv for the review!

PR #47, opened 2020-07-16
whereabouts.kubeconfig wrong in IPv6 only env

ohamada:
Server address in whereabouts.kubeconfig file has wrong format when the kubernetes service has IPv6 address. The IPv6 server address must be wrapped in '[]' in the config file.

### DIFF
--- a/script/install-cni.sh
+++ b/script/install-cni.sh
@@ -42,6 +42,12 @@ if [ -f "$SERVICE_ACCOUNT_PATH/token" ]; then
     TLS_CFG="certificate-authority-data: $(cat $KUBE_CA_FILE | base64 | tr -d '\n')"
   fi
 
+  # Kubernetes service address must be wrapped if it is IPv6 address
+  KUBERNETES_SERVICE_HOST_WRAP=$KUBERNETES_SERVICE_HOST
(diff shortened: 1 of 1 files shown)


crandles (approved):

xagent003:
Will this be merged soon?

script/install-cni.sh
```diff
@@ -42,6 +42,12 @@ if [ -f "$SERVICE_ACCOUNT_PATH/token" ]; then
     TLS_CFG="certificate-authority-data: $(cat $KUBE_CA_FILE | base64 | tr -d '\n')"
   fi
 
+  # Kubernetes service address must be wrapped if it is IPv6 address
+  KUBERNETES_SERVICE_HOST_WRAP=$KUBERNETES_SERVICE_HOST
+  if [ "$KUBERNETES_SERVICE_HOST_WRAP" != "${KUBERNETES_SERVICE_HOST_WRAP#*:[0-9a-fA-F]}" ]; then
+    KUBERNETES_SERVICE_HOST_WRAP=\[$KUBERNETES_SERVICE_HOST_WRAP\]
+  fi
+
```
mithilarun, line 50: This isn't necessary. It is okay to wrap the service IP with square braces even if it isn't IPv6. See: https://github.com/projectcalico/cni-plugin/blob/be4df4db2e47aa7378b1bdf6933724bac1f348d0/k8s-install/scripts/install-cni.sh#L104-L153

mithilarun (changes requested):

PR #49, opened 2020-07-20
Feature: Adds functionality for overlapping ranges

dougbtv:
This adds a feature to allow overlapping address ranges in Whereabouts without causing a condition where there are IP address collisions.

## Overview

One possible use case: Consider the scenario where you've originally allocated a range of `192.168.1.0/28` and you've increased the number of workloads for that range, and are now running out of address space, so you change to `192.168.1.0/24` -- with the current functionality, you would collide with all the addresses from the first range.

This feature can be overridden setting the `enable_overlapping_ranges` CNI configuration parameter to `false`. It defaults to true.

I propose that we tag a release at the commit before this feature is added.

Generally, the technique used is to create a new CRD called a `ClusterIPReservation` which is intended to reserve an IP cluster-wide (instead of the current methodology which essentially reserves an IP only address-range-wide). 

With each allocated IP address, a `ClusterIPReservation` is made, so that when allocating across ranges, a check can be made to see if a given IP address was allocated in another range.

## Limitations

While this does improve one aspect of the functionality, which prevents collisions when overlapping ranges are used, it does open the scenario for other potential problems -- such as IP address ranges being exhausted if pods are not properly cleaned up and a CNI DEL is not completely processed. Some examples of where this can happen: Systems crashing unexpectedly, or, pods being force deleted without a grace period (which can cut off a proper CNI DEL).

I'll be looking into some possibility of a cron job that can periodically clean out these ranges when the `ClusterIPReservation` are orphaned from the aforementioned scenarios... Which also exist with the `IPPool` reservations, as well. If a cronjob proves to be too complex, it may require some "bigger guns" -- such as an operator.

Also notably: This work *only* covers the Kubernetes storage backend, and the etcd storage backend for this functionality will need to be revisited.https://github.com/intel/multus-cni/blob/master/doc/quickstart.md

## Verifying the functionality

How to verify this functionality... This example uses `NetworkAttachmentDefinitions` from Multus CNI (see: [quickstart guide](https://github.com/intel/multus-cni/blob/master/doc/quickstart.md)). Alternatively, you could manipulate the on-disk CNI configurations.

We start with 4 yaml files, note that the `net-attach-defs` specify two ranges, an "original range" with a /28, and an expanded range with a /24.

```
[centos@kube-singlehost-master test]$ cat net-attach-def.yml 
metadata:
  name: original-range
spec:
  config: '{
      "cniVersion": "0.3.0",
      "type": "macvlan",
      "master": "eth0",
      "mode": "bridge",
      "ipam": {
        "type": "whereabouts",
        "datastore": "kubernetes",
        "kubernetes": { "kubeconfig": "/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig" },
        "range": "192.168.22.0/28",
        "log_file" : "/tmp/whereabouts.log",
        "log_level" : "debug"
      }
    }'
[centos@kube-singlehost-master test]$ cat net-attach-def2.yml 
metadata:
  name: expanded-range
spec:
  config: '{
      "cniVersion": "0.3.0",
      "type": "macvlan",
      "master": "eth0",
      "mode": "bridge",
      "ipam": {
        "type": "whereabouts",
        "datastore": "kubernetes",
        "kubernetes": { "kubeconfig": "/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig" },
        "range": "192.168.22.0/24",
        "log_file" : "/tmp/whereabouts.log",
        "log_level" : "debug"
      }
    }'
[centos@kube-singlehost-master test]$ kubectl create -f net-attach-def.yml 
[centos@kube-singlehost-master test]$ kubectl create -f net-attach-def2.yml 
```

Then create 2 pods, on that references the original range, and one that references the expanded range...

```
[centos@kube-singlehost-master test]$ cat original-range-pod.yml 
apiVersion: v1
kind: Pod
metadata:
  name: pod-original-range
  annotations:
    k8s.v1.cni.cncf.io/networks: original-range
spec:
  containers:
  - name: pod-original-range
    command: ["/bin/ash", "-c", "trap : TERM INT; sleep infinity & wait"]
    image: alpine
[centos@kube-singlehost-master test]$ cat expanded-range-pod.yml 
apiVersion: v1
kind: Pod
metadata:
  name: pod-exapanded-range
  annotations:
    k8s.v1.cni.cncf.io/networks: original-range
spec:
  containers:
  - name: pod-expanded-range
    command: ["/bin/ash", "-c", "trap : TERM INT; sleep infinity & wait"]
    image: alpine
[centos@kube-singlehost-master test]$ kubectl create -f original-range-pod.yml 
pod/pod-original-range created
[centos@kube-singlehost-master test]$ kubectl create -f expanded-range-pod.yml 
pod/pod-exapanded-range created
```

You should expect that unique IP addresses are assigned to each pod, e.g.:

```
[centos@kube-singlehost-master test]$ kubectl exec -it pod-original-range -- ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
3: eth0@if13: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1450 qdisc noqueue state UP 
    link/ether 8e:1c:a3:f1:43:da brd ff:ff:ff:ff:ff:ff
    inet 10.244.0.53/24 scope global eth0
       valid_lft forever preferred_lft forever
4: net1@if2: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 qdisc noqueue state UNKNOWN 
    link/ether 0a:de:70:0f:e1:f1 brd ff:ff:ff:ff:ff:ff
    inet 192.168.22.1/28 scope global net1
       valid_lft forever preferred_lft forever
[centos@kube-singlehost-master test]$ kubectl exec -it pod-exapanded-range -- ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
3: eth0@if14: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1450 qdisc noqueue state UP 
    link/ether 22:3d:15:9d:94:ee brd ff:ff:ff:ff:ff:ff
    inet 10.244.0.54/24 scope global eth0
       valid_lft forever preferred_lft forever
4: net1@if2: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 qdisc noqueue state UNKNOWN 
    link/ether aa:ef:b5:ee:0f:21 brd ff:ff:ff:ff:ff:ff
    inet 192.168.22.2/28 scope global net1
```

Note that the original range pod is assigned `192.168.22.1/28` and the pod with the expanded range is assigned `192.168.22.2/28`

(my apologies for the `exapanded` typo in this example)


pkg/storage/kubernetes.go
```diff
@@ -142,6 +142,65 @@ func (i *KubernetesIPAM) getPool(ctx context.Context, name string, iprange strin
...
+// IsAllocatedClusterWide checks for IP addresses to see if they're allocated cluster wide, for overlapping ranges
+func (i *KubernetesIPAM) IsAllocatedClusterWide(ctx context.Context, ip net.IP) (bool, error) {
+
+	// IPv6 doesn't make for valid CR names, so normalize it.
+	normalizedip := strings.ReplaceAll(fmt.Sprint(ip), "::", "-")
+
+	logging.Debugf("Clusterwide allocation check for IP: %v / namespace: %v", normalizedip, i.namespace)
+
+	clusteripres := &whereaboutsv1alpha1.ClusterIPReservation{
+		ObjectMeta: metav1.ObjectMeta{Name: normalizedip, Namespace: i.namespace},
```
crandles, line 154: If this is a cluster-scoped resource, you shouldn't be setting the namespace

pkg/storage/storage.go
```diff
@@ -28,6 +28,8 @@ type IPPool interface {
 // Store is the interface that wraps the basic IP Allocation methods on the underlying storage backend
 type Store interface {
 	GetIPPool(ctx context.Context, ipRange string) (IPPool, error)
+	IsAllocatedClusterWide(ctx context.Context, ip net.IP) (bool, error)
```
crandles, line 31: I wonder if it would make more sense for a separate `Store` implementation that uses the cluster-scoped CRD (instead of expanding on the interface).

It seems like if you are using the cluster-scoped IPReservation resource you won't need the namespace-scoped resource?

dougbtv: I think this is a great recommendation, I went ahead and changed up how we access the "cluster wide" resources, making a `ClusterStore`, so we follow a similar pattern as:

ipam to make -> Store
ipam to make -> ClusterStore

pkg/storage/kubernetes.go
dougbtv: Hey Chris, thanks for the eyes and definitely for the review, appreciate it a lot!

I went and played with it a `scope: Cluster` resource, and... I remembered that I had something in mind for a future enhancement which is this idea of "reuse of IPPools ranges across namespaces", I just put together some thoughts in #50 

So in this context... two parts:

1. I think when I go implement in openshift land, I'm going to have an easier time with having the CR namespaced.
2. I don't want to necessarily cut off design considerations because of that, but, I'm thinking in tandem with the considerations from #50 that I could also make these `ClusterIPReservations` namespaced as well, to essentially only mitigate the overlap when inside the same namespace. I would make this optional / configurable, too.

That being said... Now I think the name `ClusterIPReservation` is... Not great. Do you have any suggestions?

My first one is to make it a bit more specific and to not overload the term "cluster" and maybe something like...

`OverlappingIPRangeReservation`
`OverlappingRangeIPReservation`

or... Happy to take any input there, too.

pkg/types/types.go
```diff
@@ -41,6 +41,7 @@ type IPAMConfig struct {
 	EtcdCACertFile    string            `json:"etcd_ca_cert_file,omitempty"`
 	LogFile           string            `json:"log_file"`
 	LogLevel          string            `json:"log_level"`
+	OverlappingRanges bool              `json:"enable_overlapping_ranges"`
```
rootfs, line 44: omitempty

pkg/storage/storage.go
```diff
@@ -97,15 +114,43 @@ RETRYLOOP:
...
+				isallocated, err := clusterstore.IsAllocatedClusterWide(ctx, newip.IP)
+				if err != nil {
+					logging.Errorf("Error checking clusterwide allocation: %v", err)
+					return newip, err
+				}
+
+				if isallocated {
+					logging.Debugf("Continuing loop, IP is already allocated (possibly from another range): %v", newip)
+					// We create "dummy" records here for evaluation, but, we need to filter those out later.
+					clusterwideallocations = append(clusterwideallocations, types.IPReservation{IP: newip.IP, ContainerID: ".DuMMY."})
```
rootfs, line 130: what if `.DuMMY.` is a real containerID?

README.md
```diff
@@ -178,6 +178,14 @@ There are two optional parameters for logging, they are:
 * `log_file`: A file path to a logfile to log to.
 * `log_level`: Set the logging verbosity, from most to least: `debug`,`error`,`panic`
 
+### Overlapping Ranges
+
+The overlapping ranges feature is enabled by default, and will not allow an IP address to be re-assigned across two different ranges which overlap. However, this can be disabled.
+
+* `enable_overlapping_ranges`: *(boolean)* Checks to see if an IP has been allocated across another range before assigning it (defaults to `true`).
```
rootfs, line 185: would `enable_overlapping_ranges` be turned off if there is no VLAN ID?

pkg/storage/storage.go
dougbtv: Good catch, this is a hack! I went ahead and instead updated the `IPReservation` type to add a new property. Thanks!!

pkg/types/types.go
dougbtv: Got it, thanks!

README.md
dougbtv: Currently, no logic here to enforce or otherwise detect that...

...probably if I wanted to turn it on/off for a given vlan ID, I'd probably create an additional configuration as a `net-attach-def` and refer to it along with any vlan ID specification you have in there.

pkg/storage/kubernetes.go
dougbtv: I tried a rename, I think it might express the idea better than how I overloaded the use of the word "cluster" here, really.

pkg/storage/kubernetes.go
```diff
@@ -154,6 +154,80 @@ func (i *KubernetesIPAM) Close() error {
...
+// GetOverlappingRangeStore returns a clusterstore interface
+func (i *KubernetesIPAM) GetOverlappingRangeStore() (OverlappingRangeStore, error) {
+	return &KubernetesOverlappingRangeStore{i.client, i.containerID, i.namespace}, nil
+}
+
+// IsAllocatedInOverlappingRange checks for IP addresses to see if they're allocated cluster wide, for overlapping ranges
+func (c *KubernetesOverlappingRangeStore) IsAllocatedInOverlappingRange(ctx context.Context, ip net.IP) (bool, error) {
+
+	// IPv6 doesn't make for valid CR names, so normalize it.
+	normalizedip := strings.ReplaceAll(fmt.Sprint(ip), "::", "-")
```
s1061123, line 173: (Might be stupid q, but let me ask...) in ipv6 case, how ':' is replaced, such ash "2001::2:3:4/64"?

dougbtv: Nope, not a stupid question at all! I think this is a problem.

doc/whereabouts.cni.cncf.io_overlappingrangeipreservations.yaml
```diff
@@ -0,0 +1,53 @@
...
+          type: object
+      required:
+      - spec
+      type: object
+  version: v1alpha1
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
+status:
```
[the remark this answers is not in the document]
dougbtv: nope! Thanks for catching that. This was generated from kube-builder so it leaves in some stubbed in portions that aren't necessary.

pkg/storage/kubernetes.go
dougbtv: Added a test so I could improve this functionality.... fixed this one, also found a bug elsewhere in the code for IPv6! Thank you Tomo!!!

doc/daemonset-install.yaml
```diff
@@ -27,6 +27,7 @@ rules:
   - whereabouts.cni.cncf.io
   resources:
   - ippools
+  - overlappingrangeipreservations
```
crandles, line 30: I think this is out of date now

dougbtv: Thanks for the eye, got it!

cmd/whereabouts_test.go
```diff
@@ -161,6 +161,324 @@ var _ = Describe("Whereabouts operations", func() {
...
+		// Release the IP, second range
+		err = testutils.CmdDelWithArgs(argssecond, func() error {
+			return cmdDel(argssecond)
+		})
+		Expect(err).NotTo(HaveOccurred())
+
+	})
+
+	// 2001::2:3:4/64
+	// 	2001::2:3:0/124
```
dougbtv, line 271: delete these rems

doc/whereabouts.cni.cncf.io_clusteripreservations.yaml
```diff
@@ -0,0 +1,52 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  name: clusteripreservations.whereabouts.cni.cncf.io
```
dougbtv, line 7: delete this file.

dougbtv:
Changed my mind again, decided that I'd like to put this into a new `release-v0.4` branch so we can continue the review of functionality as is being proposed in #50 :+1:

PR #54, opened 2020-08-13
Change error message

s1061123:
This changes checks kubeconfig existence at loading IPAM config to cause appropreate error message when kubeconfig is not exist in case of kubernetes CRD mode.

Error message is:
```
You have not configured the storage engine (looks like you're using an invalid `kubernetes.kubeconfig` as stat /etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig: no such file or directory parameter in your config)
```

### DIFF
--- a/cmd/whereabouts.go
+++ b/cmd/whereabouts.go
@@ -40,8 +40,8 @@ func cmdAdd(args *skel.CmdArgs) error {
 	logging.Debugf("Beginning IPAM for ContainerID: %v", args.ContainerID)
 	newip, err := storage.IPManagement(types.Allocate, *ipamConf, args.ContainerID)
 	if err != nil {
-		logging.Errorf("Error assigning IP: %s", err)
-		return fmt.Errorf("Error assigning IP: %w", err)
+		logging.Errorf("Error at storage engine: %s", err)
+		return fmt.Errorf("Error at storage engine: %w", err)
 	}
 
 	// Determine if v4 or v6.
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -56,7 +56,7 @@ func IPManagement(mode int, ipamConf types.IPAMConfig, containerID string) (net.
 	}
 	if err != nil {
 		logging.Errorf("IPAM %s client initialization error: %v", ipamConf.Datastore, err)
-		return newip, err
+		return newip, fmt.Errorf("IPAM %s client initialization error: %v", ipamConf.Datastore, err)
 	}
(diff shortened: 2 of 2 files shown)


crandles:
I just did some manual testing and I believe we are already returning an error at this point: https://github.com/dougbtv/whereabouts/blob/158a70d5bca503785ce40cd99748782f3f4c79fa/pkg/storage/kubernetes.go#L38-L43 (`os.IsNotExist(err) == true `)
```
Error assigning IP: stat /tmp/whereabouts521554589/whereabouts.kubeconfig: no such file or directory
```

I think it might be better to decorate that error (if necessary) instead of adding an additional check to the config package.

s1061123:
@crandles thank you for your comment. If the error is caught there, the error message must be `		logging.Errorf("IPAM %s client initialization error: %v", ipamConf.Datastore, err)`

https://github.com/dougbtv/whereabouts/blob/158a70d5bca503785ce40cd99748782f3f4c79fa/pkg/storage/kubernetes.go#L38-L43

https://github.com/dougbtv/whereabouts/blob/158a70d5bca503785ce40cd99748782f3f4c79fa/pkg/storage/storage.go#L51-L60

So I'm a bit weird it...

s1061123:
Let me repro it my side.

s1061123:
@crandles Thank you for your info, I reproduced it. BTW, I have a question related it. I would like to know why we add existence check at storage package because configuration file path failure is a part of configuration validation, not storage functionality. Before `"ADD - IPAM configuration successfully read"` message, we should show the error otherwise user may misunderstand that the failure is not comes from configuration because debug message mentions `IPAM configuration successfully read`.

https://github.com/dougbtv/whereabouts/blob/158a70d5bca503785ce40cd99748782f3f4c79fa/cmd/whereabouts.go#L28-L33

dougbtv:
Hrmm, there is a little bit of semantics going on there that, it's... a storage operation to use Kubernetes (e.g. via CRs), so it's the first time where we need to actually use the configuration. So, while we did successfully load the configuration -- we haven't at this point successfully validated it, and... agreed, this does prove to have some confusion for the user. However, as long we've got the full path to the kubeconfig in the error output, it should be fairly clear.

I would be OK with extending the configuration to have validation of the kubeconfig file existing... However, if that file exists but isn't a valid kubeconfig, we'd still have the error at this point where we load it.

dougbtv:
Awesome, I'm OK to merge. I'm going to file an issue about the configuration validation, too.

Issue #51, opened 2020-07-22
CRDv2 data model


crandles:
The current data model for the IPPool CRD stores allocations for a given range in a single Kubernetes resource.

## Known Issues
- Under certain failure scenarios IP allocations may become orphaned
  - System failures, force deletion of pods (without a grace period)
  - not a specific problem with our data model: this seems to be a behavior of (kubelet?) kubernetes
- Using (and prevent the over-use of) IP addresses across multiple ranges
  - Should `127.0.0.0/22` and `127.0.0.0/23` both be able to allocate `127.0.0.1` ?
- Resource is namespace-scoped; some users may desire a cluster-scoped resource (e.g. multi-tenancy)

## Open Questions
- Is it important to support allocating the same IP in multiple ranges?
  - Could we instead move to a model where an IP is a unique resource?
  - See #49, except: why use the range resource + the cluster ip lookup? Would the cluster lookup suffice?

## Idea

TBD, come up with draft yaml spec/examples. Including some rough ideas now:

Moving to the `IP` as the `base` resource type solves for preventing duplicate IP allocations from occurring in overlapping ranges, but introduces a problem:
- How do we easily query kubernetes to determine the next available IP?
  - The `range` type is a useful bucket that we are removing

Can we solve querying IPs from a given range with using well-crafted labels? `kubectl get ip -l subnet_31=127.0.0.0` (?)
Need to determine labels and consider IPV6.

Given we have a single resource type that is associated with an allocation, we can leverage kuberentes' built-in [garbage collection](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/) capabilities to solve for the orphaned IP allocation problem.

By [configuring the pod as the owner](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#SetOwnerReference) of the IP allocation resource we can instruct Kubernetes to automatically delete the resource type when the pod is deleted. We may still clean up our resources via the `CNI DEL`, but this would serve as a fall-through to prevent IP exhaustion (With no operator, cron, or other process necessary).

## Limitations
Moving to such a model would not make it easy to support overlapping IP ranges, we may have to drop that use case.

We could potentially keep multiple CRD versions around (v1alpha1, v1beta1, etc) if this was an important use-case.

dougbtv:
+1 re: the solution for configuring the pod as the owner of the IP allocation, that's quite excellent, great suggestion.

[...]

PR #60, opened 2020-09-07
feat: add support multi-ranges

ustuzhanin:
There is three breaking changes:
1. ETCD store lock name
2. Name of kubernetes resource for pool
3. Replace ip offset in CRD to IP strings

### DIFF
--- a/pkg/allocate/allocate.go
+++ b/pkg/allocate/allocate.go
@@ -20,13 +20,21 @@ func (a AssignmentError) Error() string {
 	return fmt.Sprintf("Could not allocate IP in range: ip: %v / - %v / range: %#v", a.firstIP, a.lastIP, a.ipnet)
 }
 
+type DeallocateError struct {
+	contaiterID string
+}
+
+func (a DeallocateError) Error() string {
+	return fmt.Sprintf("Could not Deallocate IP for: %s", a.contaiterID)
+}
+
 // AssignIP assigns an IP using a range and a reserve list.
-func AssignIP(ipamConf types.IPAMConfig, reservelist []types.IPReservation, containerID string) (net.IPNet, []types.IPReservation, error) {
+func AssignIP(ipRange types.IPRange, reservelist []types.IPReservation, containerID string) (net.IPNet, []types.IPReservation, error) {
 
 	// Setup the basics here.
-	_, ipnet, _ := net.ParseCIDR(ipamConf.Range)
+	_, ipnet, _ := net.ParseCIDR(ipRange.Range)
 
-	newip, updatedreservelist, err := IterateForAssignment(*ipnet, ipamConf.RangeStart, ipamConf.RangeEnd, reservelist, ipamConf.OmitRanges, containerID)
+	newip, updatedreservelist, err := IterateForAssignment(*ipnet, ipRange.RangeStart, ipRange.RangeEnd, reservelist, ipRange.OmitRanges, containerID)
 	if err != nil {
 		return net.IPNet{}, nil, err
 	}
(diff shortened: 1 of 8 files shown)


ustuzhanin:
@dougbtv

dougbtv:
Thanks a ton for the contribution! This will take a bit of time to dig through and think through for sure. So give me a little time, but, thanks a bunch for putting this together, know that I appreciate it.

pkg/storage/storage.go
```diff
@@ -66,9 +66,43 @@ func IPManagement(mode int, ipamConf types.IPAMConfig, containerID string) (net.
...
+		logging.Debugf("try ipRange:%+v", ipRange)
+		newip, err = IPManagementRange(ctx, mode, ipRange, ipam, containerID)
+		logging.Debugf("result allocate in range %+v IP:%+v err:%v", ipRange, newip, err)
+		if err == nil {
+			gateway = ipRange.Gateway
+			break
+
+		} else {
+			switch mode {
+			case 0:
```
crandles, line 83: can you use the constants `types.Allocate` and `types.Deallocate` instead of 0/1?

https://github.com/dougbtv/whereabouts/blob/master/pkg/types/types.go

pkg/storage/kubernetes.go
```diff
@@ -58,7 +52,7 @@ func NewKubernetesIPAM(containerID string, ipamConf whereaboutstypes.IPAMConfig)
 	if err != nil {
 		return nil, err
 	}
-	return &KubernetesIPAM{c, ipamConf, containerID, namespace, DatastoreRetries}, nil
+	return &KubernetesIPAM{c, ipamConf, containerID, namespace, ipamConf.Name, DatastoreRetries}, nil
```
crandles, line 55: Interesting, so instead of using the ip range as the resource name, this is utilizing the `"name"` of the CNI config (e.g. `mynet` from the tests). 

I'm not against this, we're rethinking the pool resource type in #51; this is a step in that direction I believe.

pkg/storage/storage.go
ustuzhanin: fixed

ustuzhanin:
And what will be the solution? We have already implemented it - it works

ustuzhanin:
Hi :) We have been using this patch in production since October 20th. It works. Will there be movements?

dionysius:
Allow me to hop in. We'd really like to see this feature. This all started when I asked for this feature in https://github.com/openshift/whereabouts-cni/issues/13.

Allow me to copy the issue description here:

-----

There seems to be no support for multiple ranges (or at least not documented). It seems this *very crucial* part of the host-local plugin has not been picked up.

### Use case
A /24 range is used up, since the cluster grew and grew. You could want to go /23, but the next /24 range is already reserved or you can't use the same gateway as the *actual* network is a /24 and the other half wouldn't be able to reach the gateway. So it's **technically not possible** to extend the range!

### Proposal
Introduce a `ranges` property which accepts a slice of `range-item`. A `range-item` consists of: `range`, `range_start`, `range_end`, `exclude`, `gateway` and as well the properties from the static plugin `dns`, `routes` and `addresses` as you already do.

So basically the the same properties you have as right now, but allowing to put them into a slice.

### Compatibility
Parsing of the current format would still be valid. Operational code would be changed to work with the `ranges` slice. After parsing the config the values of these fields created a new `range-item` and are added to the `ranges` slice. And since we changed the operational code to to work with the `ranges` slice it is already ready to be applied.

### Otherwise
Once the range is used up, you have to migrate **everything** to a completely new cluster with a new but bigger range, but also still limited until you find yourself in this situation again.

e0ne (changes requested):

pkg/allocate/allocate.go
```diff
@@ -35,7 +43,7 @@ func AssignIP(ipamConf types.IPAMConfig, reservelist []types.IPReservation, cont
 }
 
 // DeallocateIP assigns an IP using a range and a reserve list.
-func DeallocateIP(iprange string, reservelist []types.IPReservation, containerID string) ([]types.IPReservation, error) {
+func DeallocateIP(_ types.IPRange, reservelist []types.IPReservation, containerID string) ([]types.IPReservation, error) {
```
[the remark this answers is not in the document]
ustuzhanin: fixed

ustuzhanin:
i am merge upstream, conflicts reloved.
need any more?

PR #65, opened 2020-09-17
update to kubebuilder v2.3.1 tooling, controller-gen v0.4.0, and update CRD apiVersion to v1

crandles:
Updating the CRD tooling to the latest kubebuilder / controller-gen / controller-runtime releases in preparation for #51.

One side affect of this change: The `apiVersion` is now `apiextensions.k8s.io/v1` in the CRDs, which requires Kubernetes 1.16.

I was able to continue executing controller-gen v0.2.0 against the codebase to generate the CRDs for older versions. I need to confirm how this works in practice on a few versions of Kubernetes, this may be broken at the moment for the pre-v1.16 v1beta1 users.

## TODO
- [x] test with multiple kubernetes versions
- [ ] test on real v1.14 / 1.16 clusters


crandles:
Tests are passing locally. I'm going to test this on a few clusters next week. I can't tell if this is passing in travis yet, the job doesn't look to have triggered.

dougbtv:
Sounds like a solid plan. @crandles feel free to ding me on kube slack if your tests are done and I haven't circled back around for a review. I gave a skim and I'm good with these tooling changes, so... probably nothing super big to change, but... Love the real-world-cluster test.

...Also I gotta switch this up to GitHub actions (as opposed to Travis) at some point. I had an intern give it a first pass, I gotta dig up his work too and maybe I can go from there.

hack/install-kubebuilder-tools.sh
```diff
@@ -1,15 +1,31 @@
...
+# install kubebuilder-1.14 tools
 containerID=$(docker create gcr.io/kubebuilder/thirdparty-linux:1.14.1)
 docker cp ${containerID}:/kubebuilder_linux_amd64.tar.gz ./kubebuilder_linux_amd64.tar.gz
 docker rm ${containerID}
 tar -xzvf kubebuilder_linux_amd64.tar.gz
 rm kubebuilder_linux_amd64.tar.gz
-mv kubebuilder/bin/* bin/
-rm -rf kubebuilder/
+mv kubebuilder kubebuilder-1.14
+chmod +x kubebuilder-1.14/bin/
```
zrhmn, line 20: There should probably be a `-R` somewhere (if the intention is to mark the contents of `kubebuilder-1.14/bin` executable).

crandles:
I'm going to try to wrap this up by next week, and finally put some code towards the new data model.

PR #72, opened 2020-11-12
Excluded subnet handling for ipv6

shivakumargn:
Excluded subnets was handled for ipv4. This fix handles ipv6 as well.

Fixes dougbtv/whereabouts#71


dougbtv:
@shivakumargn thanks for the fix! I'm going to merge into a develop branch and add a couple unit tests, thanks!

PR #73, opened 2020-11-12
Includes fix for excluding IPv6 ranges as well as unit test

dougbtv:


dougbtv:
@shivakumargn mind taking a look as well? I'll likely merge it assuming it passes / builds OK. Thanks again.

PR #74, opened 2020-11-12
Removes error when deallocating IP errors out, instead just warns.

### LINKED ISSUE #70: Pod gets stuck in ContainerCreating state if first attempt of CNI ADD fails (Multus CNI, Whereabouts IPAM)
See: https://github.com/intel/multus-cni/issues/578
Also: https://github.com/kubernetes/kubernetes/issues/96200

I have a possible work-around, my thought was "Is it really an error when we can't deallocate an IP? I mean... it's warning worthy, but, in the case where a pod is created, can't get assigned an IP, and then a CNI DEL comes along... it's just a warning to say I can't do that" and when I do this on CNI DEL, I instead just log an error and exit 0.

Have an experimental change in my `stuck-created` branch in this commit: https://github.com/dougbtv/whereabouts/commit/9769d4b68292eb75f4afadd9b1e0344d48e70516

dougbtv:
@crandles any thoughts on this one?

I've gotten a few reports about a pod being in stuck creating when a range gets exhausted.

* https://github.com/kubernetes/kubernetes/issues/96200
* https://github.com/intel/multus-cni/issues/578
* https://github.com/dougbtv/whereabouts/issues/70

Overall, I think it's generally more correct... The gist is we get into a kind of loop when we have a situation where there's CNI ADD "no IP address there.... error out" and then CNI DEL "...error, because I never allocated an IP". I think CNI DEL in that case should instead be a warning, and not exit with an error.

Fixes #70

### DIFF
--- a/cmd/whereabouts.go
+++ b/cmd/whereabouts.go
@@ -79,8 +79,8 @@ func cmdDel(args *skel.CmdArgs) error {
 
 	_, err = storage.IPManagement(types.Deallocate, *ipamConf, args.ContainerID)
 	if err != nil {
-		logging.Errorf("Error deallocating IP: %s", err)
-		return fmt.Errorf("Error deallocating IP: %s", err)
+		logging.Verbosef("WARNING: Problem deallocating IP: %s", err)
+		// return fmt.Errorf("Error deallocating IP: %s", err)
 	}
 
 	return nil


dougbtv:
After a discussion with Dan Williams, I'm feeling fairly positive about the change to only warn on exit. I think I'm going to move ahead with a merge.

@TothFerenc @shivakumargn

crandles:
LGTM

TothFerenc:
@dougbtv : Thanks! Do you plan a new release anytime soon? :)

shivakumargn:
LGTM

nelsonpraveen:
@dougbtv  - Is there any image available with the fix ? I tried using the dougbtv/whereabouts:experimental image and still facing issue.
Log message:
2021-02-20T17:39:51Z [error] Error deallocating IP: Did not find reserved IP for container 6061894b51c5c616385574d949c59493575981a1bc8da105a06827897adfb1fd2021-02-20T17:39:51Z [verbose] WARNING: Problem deallocating IP: Did not find reserved IP for container 6061894b51c5c616385574d949c59493575981a1bc8da105a06827897adfb1fd

Could it be due to earlier IPs still stored in the etcd and whereabouts unable to clear them ? Any pointers would help. Thanks.

shivakumargn:
You are probably facing https://github.com/dougbtv/whereabouts/issues/75 or a variant of it.

Issue #75, opened 2020-11-13
Node crash results in IPs not getting deallocated


shivakumargn:
If a kubernetes node crashes, whereabouts assigned IPs are not deallocated. New pods do not get the IPs.

A monitoring capability is currently a missing capability.

shivakumargn:
I have noticed that if the crashed node is brought back up (even after a long time), the IP gets cleared.

In the absence of a clean solution, a temperory resolution can be manually clear the etcd. But due to the behavior of IP getting cleared after the crashed node comes up, I am not sure if cron job solution is without side-effects as below:

1. Node crashes
2. Cron job clears the IP
3. IP is assigned to new pod
4. Crashed node comes up
5. IP gets cleared (because of crashed node coming up) --> but IP was already assigned to new pod on different node

dougbtv:
Thanks for the report. Agreed, some monitoring and cleanup routines are necessary. I think it's possible there needs to be a Whereabouts controller that can perform the cleanup after some given period of time -- or... The cron job actually sounds very pragmatic.

I wonder how the IPs get deallocated when the node comes up, that's a mystery to me. Is it that a CNI DEL is finally called when the node comes up? Hrmm, I'm unsure.

crandles:
I suspect when the node starts, kubelet likely cleans up the state associated with the node as it is able to delete pods (that were likely in an unknown state).

shivakumargn:
> I wonder how the IPs get deallocated when the node comes up, that's a mystery to me. Is it that a CNI DEL is finally called when the node comes up? Hrmm, I'm unsure.

Right now I do not have a setup to try. Once I am able to will share the details.

xagent003:
Hi @dougbtv Is there any update on this? Is the new ip-reconciler intended as a complete fix for this?

PR #86, opened 2021-02-02
Enrich the IP allocation data structure to store the Pod reference

TothFerenc:
As it is discussed in #75, some kind of cleaner mechanism is needed to release non-used IPs.

Whereabouts stores the ContainerID for each allocated IP. In case of Docker runtime that ContainerID is the ID of the **infra** container. The problem is that K8s does not register this infra container ID in the Pod descriptor, so later on we cannot determine which Pod owns which allocated IP (not considering the method about executing `docker ps` on every node to find out).

This PR would be the first step to resolve this issue by enriching the IPAllocation structure to store a Pod reference string (in `namespace/name` format). So later on a "cleaner" controller/script can walk through the IP allocation map and check whether each referenced Pod is still running or not. If a Pod is missing, it is safe to release its IP.

New IPPool example:
```
[..]
  spec:
    allocations:
      "1":
        id: 0c85206ce40fd429c62d1499c91f0fd5cb183fd610db3c4c458f167eaa0ee190
        podref: test/test-648b75d6fc-f56cp
      "2":
        id: 6442798c086abdad4c636b7c8e51f427511076a250e5434889b675c9e60bca06
        podref: test/test-648b75d6fc-sdbbm
      "3":
        id: c37586753ac82a69d7afbe73602020e7ecf4a93c666589d15cfb0108012e24ee
        podref: test/test-648b75d6fc-bhdjd
[..]
```

This new `podref` field is a description only for IPPool viewers. Its content is not considered during IP allocation/deallocation logic.

### DIFF
--- a/cmd/whereabouts.go
+++ b/cmd/whereabouts.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/containernetworking/cni/pkg/skel"
 	cnitypes "github.com/containernetworking/cni/pkg/types"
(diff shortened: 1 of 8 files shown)


TothFerenc:
@dougbtv : What's your opinion about this PR?

dougbtv:
@TothFerenc I haven't forgotten about you! I'm sorry there's a delay here. I'm definitely in favor of this change and appreciate it greatly. Let me give it a shot in my environment we can proceed with integrating this change. Thanks!

dougbtv:
@TothFerenc I'm ready to move ahead with this PR, would you mind giving it a rebase for us? Thanks!

TothFerenc:
@dougbtv : PR has been rebased.

dougbtv:
Thanks again Ferenc, merged!

Issue #84, opened 2021-01-22
When the first hextet is shortened it can cause a crash.


dougbtv:
To test this, add to allocate_test.go:

```
	It("creates an IPv6 range when the first hextets's zeroes have been shortened", func() {

		ip, ipnet, err := net.ParseCIDR("fd:db8:abcd:0012::0/96")
		ip, _ = AddressRange(ipnet)

		Expect(err).NotTo(HaveOccurred())

		firstip, lastip, err := GetIPRange(net.ParseIP(ip.String()), *ipnet)
		Expect(err).NotTo(HaveOccurred())

		Expect(fmt.Sprint(firstip)).To(Equal("fd:db8:abcd:12::1"))
		Expect(fmt.Sprint(lastip)).To(Equal("fd:db8:abcd:12::ffff:ffff"))

	})

```

Notice that it's `fd` in the first octet, if you can change that to `2001` or `fd00`, that will work.

Currently, you'll get an error that looks like:

```
panic: runtime error: index out of range [16] with length 16

goroutine 1 [running]:
github.com/dougbtv/whereabouts/pkg/allocate.BigIntToIP(0xc0002e6700, 0xc0000ebb30, 0x2, 0x6, 0x4, 0x4, 0xc0002c3200)
\t/go/src/github.com/dougbtv/whereabouts/pkg/allocate/allocate.go:235 +0x15e
```

I'm starting to put some traces in to see where the logic breaks down for these.

dougbtv:
Also, of note is that this is reserved address space:

https://www.iana.org/assignments/ipv6-address-space/ipv6-address-space.xhtml

https://tools.ietf.org/html/rfc3513#section-2.5:

PR #88, opened 2021-02-15
[WIP] Fix for IP addresses which start with any number of bytes with a zero value

dougbtv:
Tomo also has a fix together that may be a more elegant solution that doesn't rely on bigint. However, this implementation is currently working.

For #84


dougbtv:
Closing in favor of #89

PR #89, opened 2021-02-16
Remove bigint code to fix starts-with-zero-ipv6-problem

### LINKED ISSUE #84: When the first hextet is shortened it can cause a crash.

s1061123:
Fix #84
This code removes bigint from pkg/allocate because it may causes various problem.

### DIFF
--- a/cmd/whereabouts_test.go
+++ b/cmd/whereabouts_test.go
@@ -162,6 +162,16 @@ var _ = Describe("Whereabouts operations", func() {
 		AllocateAndReleaseAddressesTest(ipVersion, ipRange, ipGateway, []string{expectedAddress}, whereaboutstypes.DatastoreKubernetes)
 	})
 
+	It("allocates and releases an IPv6 address with left-hand zeroes on ADD/DEL with a Kubernetes backend", func() {
+
+		ipVersion := "6"
+		ipRange := "fd::1/116"
+		ipGateway := "fd::f:1"
+		expectedAddress := "fd::1/116"
+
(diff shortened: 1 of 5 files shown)


dougbtv (approved):
Tomo, looks great. Thanks for working out how to accomplish this without bigint. And all the beautiful bitwise logic!! :100:

dougbtv:
Travis build is failing due to something choking on coveralls with: `{"message":"service_job_id (759238235) must be unique for Travis Jobs not supplying a Coveralls Repo Token","error":true}`

I'll give it another try tomorrow, but if it still chokes on this, I may move forward with it, it's passing all other tests.

PR #96, opened 2021-02-27
feat: add arm64 dockerfile and add github action steps for new arch

macneib:
**Problem:**  

- When testing with k3s cluster and pi4 targets (arm64), this where-aboot project does not work in Canada. :canada:

**Key Features:**

- Provides `Dockerfile.arm64` to create arm64 compatible docker images
- Github actions to build image

**Notes:**
- Tested local docker build only, not yet deployed against k3s cluster.


dougbtv (approved):
Looks great, thank you Blair @macneib !

+10000000000 points for observing the Canadian limitations :canada: I will crack open a bag of all dressed chips to celebrate the successful testing in the great north!

PR #113, opened 2021-06-24
Add leader election

dougbtv:
Uses `k8s.io/client-go/tools/leaderelection` to add leader election to choose a leader during CRD read/write in order to properly handle concurrency before ippools are calculated and modified.

### DIFF
--- a/pkg/storage/etcd.go
+++ b/pkg/storage/etcd.go
@@ -10,6 +10,8 @@ import (
 	"github.com/coreos/etcd/clientv3"
 	"github.com/coreos/etcd/clientv3/concurrency"
 	"github.com/coreos/etcd/pkg/transport"
+	"github.com/dougbtv/whereabouts/pkg/allocate"
+	"github.com/dougbtv/whereabouts/pkg/logging"
 	"github.com/dougbtv/whereabouts/pkg/types"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/storage/kubernetes.go
```diff
@@ -227,3 +239,197 @@ func (t *temporaryError) Temporary() bool {
...
+	})
+	if err != nil {
+		logging.Errorf("Failed to create leader elector: %v", err)
+		return nil, leaderOK, deposed
+	}
+	return le, leaderOK, deposed
+}
+
+/*
+ */
```
s1061123, line 291: Just remove it.

pkg/storage/kubernetes.go
```diff
@@ -227,3 +239,197 @@ func (t *temporaryError) Temporary() bool {
...
+			Identity: fmt.Sprintf("%s/%s", podNamespace, podID),
+		},
+	}
+
+	// Make the leader elector, ready to be used in the Workgroup.
+	le, err := leaderelection.NewLeaderElector(leaderelection.LeaderElectionConfig{
+		Lock:          rl,
+		LeaseDuration: 1500 * time.Millisecond,
+		RenewDeadline: 1000 * time.Millisecond,
+		RetryPeriod:   500 * time.Millisecond,
```
s1061123, line 269: Can we make it configurable from CNI json config? It might be key for performance tuning.

s1061123 (changes requested):

pkg/storage/kubernetes.go
```diff
@@ -227,3 +239,197 @@ func (t *temporaryError) Temporary() bool {
...
+	//log.WithField("context", "leaderelection")
+	// leaderOK will block gRPC startup until it's closed.
+	leaderOK := make(chan struct{})
+	// deposed is closed by the leader election callback when
+	// we are deposed as leader so that we can clean up.
+	deposed := make(chan struct{})
+
+	var rl = &resourcelock.LeaseLock{
+		LeaseMeta: metav1.ObjectMeta{
+			Name:      "whereabouts",
```
crandles, line 255: you could potentially use the name of the pool here, so that allocations in other pools are not also blocked

s1061123: I suppose we should use leaselock to lock whole whereabouts, not per ip pool basis because Kubernetes/etcd does not support fine grain lock (because etcd is not DB, just KVS).

pkg/storage/etcd.go
```diff
@@ -137,3 +139,90 @@ func (p *ETCDIPPool) Update(ctx context.Context, reservations []types.IPReservat
...
+
+	var ipam Store
+	var pool IPPool
+	var err error
+	switch ipamConf.Datastore {
+	case types.DatastoreETCD:
+		ipam, err = NewETCDIPAM(ipamConf)
+	case types.DatastoreKubernetes:
+		ipam, err = NewKubernetesIPAM(containerID, ipamConf)
+	}
```
maiqueb, line 164: It seems strange to me to find this switch in the `etcd` implementation. Shouldn't the correct config be passed as a parameter to the implementations ?

pkg/storage/kubernetes.go
dougbtv: Added changes for configurable parameters for lease/renew/retry durations

dougbtv:
New commit addresses configurability of the duration parameters, adds defaults based on original code

pkg/storage/kubernetes.go
```diff
@@ -227,3 +239,197 @@ func (t *temporaryError) Temporary() bool {
...
+	close(stopM)
+	logging.Debugf("IPManagementKubernetes: %v, %v", newip, err)
+	return newip, err
+}
+
+// IPManagementKubernetesUpdate manages k8s updates
+func IPManagementKubernetesUpdate(mode int, ipam *KubernetesIPAM, ipamConf whereaboutstypes.IPAMConfig, containerID string, podRef string) (net.IPNet, error) {
+	logging.Debugf("IPManagement -- mode: %v / containerID: %v / podRef: %v", mode, containerID, podRef)
+
+	now := time.Now()
```
s1061123, line 363: Please remove this because it is only for debug purpose.

pkg/storage/kubernetes.go
```diff
@@ -227,3 +239,197 @@ func (t *temporaryError) Temporary() bool {
...
+			if e, ok := err.(temporary); ok && e.Temporary() {
+				continue
+			}
+			break RETRYLOOP
+		}
+		break RETRYLOOP
+	}
+
+	//XXX
+	logging.Debugf("Took: %vms\n", time.Since(now).Milliseconds())
```
s1061123, line 433: Please remove this because it is only for debug purpose.

s1061123 (changes requested):

maiqueb (changes requested):
It misses `hack/update-deps.sh` .

pkg/storage/kubernetes.go
maiqueb: Right! And this line causes the openshift img build failure, since it uses golang-1.12 - `time.Milliseconds` is available on golang >= 1.13 as per https://golang.org/pkg/time/#Duration.Milliseconds

PR #114, opened 2021-06-29
replace patch with update api to rewrite IPPool

pperiyasamy:
as we noticied patch operation doesn't provide concurrency safety (refer this [link](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/)), replace it with update api to make ip assignment to be atomic.

`TODO: large scale verification is still pending`

### DIFF
--- a/cmd/whereabouts.go
+++ b/cmd/whereabouts.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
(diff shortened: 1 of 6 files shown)


dougbtv:
thanks for the contribution and your contribution to the final outcome for the write-locking issues we had.

PR #115, opened 2021-06-29
use IPPoolLock CR for rewriting IPPool

pperiyasamy:
another approach (as opposed to #114 and #113 ) to solve concurrency issue with IP assignment. this introduces IPPoolLock for ip range and this must be acquired before allocating/deallocating IP address for a pod and released once operation is complete.

`TODO: large scale verification is still pending`


s1061123 (changes requested):
This introduces new CRD hence you need to consider about version migration. What is happen if user runs this new whereabouts without IPPool CRD?

rthakur-est:
Instead of CRD, lease object (https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/lease-v1/) can be used as a lock

pkg/storage/kubernetes.go
```diff
@@ -165,49 +206,11 @@ func (p *KubernetesIPPool) Allocations() []whereaboutstypes.IPReservation {
...
-	}
-	ops = append(ops, patch...)
-	patchData, err := json.Marshal(ops)
-	if err != nil {
-		return err
-	}
-
-	// apply the patch
-	err = p.client.Patch(ctx, orig, client.ConstantPatch(types.JSONPatchType, patchData))
+	err := p.client.Update(ctx, p.pool)
```
rthakur-est, line 211: With lock based approach, it is expected that only one client will update the ip pool at a time. So, json patch api should continue to be used instead of Update api.

pkg/storage/kubernetes.go
```diff
@@ -57,6 +58,31 @@ func NewKubernetesIPAM(containerID string, ipamConf whereaboutstypes.IPAMConfig)
...
 	}
+	lockName := getNormalizedIpRangeStr(ipamConf.Range)
+	poolLock := &whereaboutsv1alpha1.IPPoolLock{
+		ObjectMeta: metav1.ObjectMeta{Name: lockName, Namespace: namespace},
+	}
+	// try creating ip pool lock, retry (if already exists) until it succeeds
+	// this makes serial access to ip pool across cluster wide and this might decrease
+	// the load considerably on k8 api server because of too many update calls, this can
+	// happen due to retries upon status conflict errors
+	for {
```
rthakur-est, line 69: Before calling Create(), the client can do a Get() for the resource. A successful Get() will indicate that the lock is being held by someone else and it should wait for sometime and retry.

pkg/storage/kubernetes.go
```diff
@@ -57,6 +58,31 @@ func NewKubernetesIPAM(containerID string, ipamConf whereaboutstypes.IPAMConfig)
...
+		err = c.Create(ctx, poolLock)
+		if err != nil {
+			if errors.IsAlreadyExists(err) {
+				interval, _ := rand.Int(rand.Reader, big.NewInt(1000))
```
rthakur-est, line 73: range of sleep interval of 1000ms seems large for in cluster clients as per this link:
https://pkg.go.dev/k8s.io/client-go/util/retry?utm_source=godoc#RetryOnConflict
should this be changed to range of 100ms?

dougbtv:
thanks for the contribution and your contribution to the final outcome for the write-locking issues we had.

PR #116, opened 2021-06-30
ippools schema: add podref

maiqueb:
PR #86 added support for a `podRef` field in the `IPReservation`.
Unfortunately, the IPPools CR was not updated, which causes the
whereabouts installation to miss it.

This commit adds the `podRef` attribute to the CR, as optional.


dougbtv:
SHIP IT! :shipit: Thanks Miguel!

PR #119, opened 2021-07-06
return context timeout error for deallocate

pperiyasamy:
return context timeout error back to cni/kubelet in case of pod delete, this would allow kubelet to retry for deallocate and avoids stale ip pool entries in the datastore after pod delete.

context timeout errors (due to default 10s request timeout) are frequently seen in highly scaled setup.

Signed-off-by: Periyasamy Palanisamy <periyasamy.palanisamy@est.tech>

### DIFF
--- a/cmd/whereabouts.go
+++ b/cmd/whereabouts.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"strings"
 
@@ -42,9 +43,6 @@ func cmdAdd(args *skel.CmdArgs) error {
 	newip, err := storage.IPManagement(types.Allocate, *ipamConf, args.ContainerID, getPodRef(args.Args))
 	if err != nil {
 		logging.Errorf("Error at storage engine: %s", err)
-		if e, ok := err.(storage.Temporary); ok && e.Temporary() {
-			err = e.GetCause()
-		}
 		return fmt.Errorf("error at storage engine: %w", err)
 	}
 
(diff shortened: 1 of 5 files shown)


s1061123:
In https://github.com/containernetworking/cni/blob/spec-v0.4.0/SPEC.md, there is following section and your change violates the CNI specification because your code may return error in DEL action.

> Plugins should generally complete a DEL action without error even if some resources are missing. For example, an IPAM plugin should generally release an IP allocation and return success even if the container network namespace no longer exists, unless that network namespace is critical for IPAM management. While DHCP may usually send a 'release' message on the container network interface, since DHCP leases have a lifetime this release action would not be considered critical and no error should be returned. For another example, the bridge plugin should delegate the DEL action to the IPAM plugin and clean up its own resources (if present) even if the container network namespace and/or container network interface no longer exist.

JanScheurich:
Not returning an error means whereabouts accumulates stale IP addresses in its IP pools. It would require an off-line audit function to find and delete such stale IPs. 
I read the CNI specification a bit differently: The CNI should not return an error at DEL if (part of) the resources to be deleted are missing already. But in our case, the actual deletion fails (typically due to too many collisions on the storage back-end).

pperiyasamy:
> Plugins should generally complete a DEL action without error even if some resources are missing.

@s1061123 thanks for this cni spec pointer! good info. just found an issue with ovs-cni due to missing net ns directory. it needs fixing there.
but in this case, as jan mentioned, whereabouts needs to propagate context timeout error back to cni/kubelet when ip entry is not removed from datastore. otherwise having `error` return type for `cmdDel` doesn't make any sense at all.

dougbtv:
Any chance you could rebase and see if this is still an issue, thanks!

bpickard22:
no update for 3 years, closing

PR #121, opened 2021-07-08
build: update openshift build image to golang 1.15

maiqueb:
This new Dockerfile for openshift was inspired by the multus one,
available at [0].

[0] - https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/deployments/Dockerfile.openshift#L2


dougbtv:
awesome! thanks! lgtm

PR #123, opened 2021-07-08
Use k8snetworkplumbingwg repo instead of dougbtv in sources

e0ne:


dougbtv:
This is great btw. Sorry for the delay. I have to prioritize a few merges first, and this might require a rebase after, but... I truly appreciate it. Thanks!

e0ne:
Thank, @dougbtv! Feel free to ping me when rebase will be needed

moshe010 (approved):

dougbtv (changes requested):
Ok, the dependent PRs have been merged, I think it's time for a rebase (or, I guess alternatively you could use the same replace technique against master @ HEAD.) Thanks again!

e0ne:
@dougbtv code is rebased against the latest master

gaby:
@dougbtv Can you merge this? We would like to use the Dockerfile without having to make changes when doing a git clone.

e0ne:
thanks for the review @dougbtv

Issue #110, opened 2021-05-25
Duplicate IP addresses at scale: Possible read/write locking problem?


dougbtv:
@pperiyasamy has reported that they're seeing duplicate IP addresses on a large k8s cluster.

He's pointed out that a lock may not be acquired like it is in the etcd implementation (the same problem does not appear when using the etcd backend)

https://github.com/k8snetworkplumbingwg/whereabouts/blob/master/pkg/storage/storage.go#L82-L108

Does this need to have a further locking mechanism?

How to test for the duplicate IP problem with this scaled up concurrency?

crandles:
The Kubernetes CRD backend does not use locks, but does use what was at the time a recommended alternative: https://github.com/k8snetworkplumbingwg/whereabouts/blob/5e8cafde59fc910f032df1e58997751920957ca4/pkg/storage/kubernetes.go#L182-L199 -- if I can find a reference I'll add a few links. IMO, using annotations / resource versions and the jsonpatch "transaction" matched closely to how locks are often implemented with etcd.

The intent here was to use JSON patch for the update operation, with added tests to:
- Only allow the update operation to occur on the expected resource version
  - this _should_ prevent two operations from being written at the same time
- ensure patched paths (adds) are empty
  - this seemed like an extra safeguard that was good to add

Perhaps something is going wrong with the patch, or how the tests were assumed to work.

How many IPs are in use when this occurs? Could we be hitting our upper bound? I know that we have one, but I have not looked into it before.

I am a few versions behind, and only use a few IPs so may not be hitting this edge case.

JanScheurich:
We see this issue massively (~5 duplicates per IP address) already with a 100 pod deployment in a cluster of 80 node when using a huge /8 IPv4 range. So lack of available IPs we can rule out. In the kubelet logs we observed a huge number of where about retries due to failed patch operations. Apparently the concurrency protection of the patch is working at times but does not prevent duplicate assignments fully. I can't rule out that we hit the upper limit of retries, but If I look at the code that should lead to IP address assignment failure. We don't see that.

pperiyasamy:
> The intent here was to use JSON patch for the update operation, with added tests to:
> 
> * Only allow the update operation to occur on the expected resource version
>   
>   * this _should_ prevent two operations from being written at the same time (and retry logic should allow eventual success)
> * ensure patched paths (adds) are empty
>   
>   * this seemed like an extra safeguard that was good to add
> 
[...]

PR #127, opened 2021-07-30
Use lease for distributed locking with k8s store

rthakur-est:
This change adds k8s lease object that is used as a distributed lock
for IP allocation and deallocation with k8s datastore.
Each request first tries to acquire the lease before doing
allocation/deallocation and updates itself as the owner of the lease.
If lease is held by someone else or conflict occurs while taking
ownership, the client sleeps for a random time in the interval of
configured backoff.

No duplicate IPs/stale IPs are seen with the patch.
Time taken for deployment/undeployment of 500 pods on 100 node setup:

- Deployment - 4 min 34 seconds
- Undeployment - 5 min 49 seconds

#110 

Signed-off-by: Ravindra Thakur <ravindra.nath.thakur@est.tech>


maiqueb (commented):
Hi! Thanks for the PR. 

Would you @rthakur-est help me understand what is the problem you're trying to solve (I'm assuming that it is that leader election is too slow at scale... better be sure :) ) and why this is better than what we currently have ? 

Ideally, I'd like to see both a small explanation and a script (or test!) so I could get some figures and compare this with the current implementation.

It's hard accepting this without some proof that this indeed performs better.

I'll try to review this code never the less.

rthakur-est:
Hi @maiqueb 

This patch is there to improve over the leader election version which runs too slow with scale as you have already pointed out.
We have tried this patch on a 100 node system with backoff of 2500 ms and following was the time duration -
a. deployment - 4 min 34 seconds
b. undeployment - 5 min 49 seconds

Also, there were no duplicate IP assignments and stale IPs seen during deployment/undeployment respectively.

maiqueb:
> Hi @maiqueb
> 
> This patch is there to improve over the leader election version which runs too slow with scale as you have already pointed out.
> We have tried this patch on a 100 node system with backoff of 2500 ms and following was the time duration for 500 pods -
> a. deployment - 4 min 34 seconds
> b. undeployment - 5 min 49 seconds

Thanks for the figures ! Would kindly provide the figures for this scenario **without** this PR ?

Please update this PR's description with this (really helpful) info.

> 
> Also, there were no duplicate IP assignments and stale IPs seen during deployment/undeployment respectively.

Cool.

rthakur-est:
Please see results published by Jan here for different patches including the leader election version: https://github.com/k8snetworkplumbingwg/whereabouts/issues/110#issuecomment-880050570

maiqueb:
> Please see results published by Jan here for different patches including the leader election version: [#110 (comment)](https://github.com/k8snetworkplumbingwg/whereabouts/issues/110#issuecomment-880050570)

Wow! Wasn't aware of that. Thanks for pointing it out !!

JanScheurich:
@rthakur-est: It's great that your lease lock approach succesfully prevents both duplicate and stale IPs. Could you, for reference and comparison, also provide measurements on the same 100 node setup for a) the original faulty SW baseline (prior to the leader election fix) and report the number of duplicates and stales and b) the etcd storage backend?

pkg/storage/kubernetes/ipam.go
```diff
@@ -2,31 +2,168 @@ package kubernetes
...
+		return fmt.Errorf("failed to acquire lease %s after waiting, client: %s",
+			lease.Name, clientID)
+	}
+	logging.Debugf("lease %s acquired, client: %s", lease.Name, clientID)
+	return nil
+}
+
+func releaseLease(ipam *KubernetesIPAM, ipamConf *whereaboutstypes.IPAMConfig) error {
+	leaseName := strings.ReplaceAll(ipamConf.Range, ":", "-")
+	leaseName = strings.ReplaceAll(leaseName, "/", "-")
```
maiqueb, line 136: Here you could re-use the helpers I've mentioned in the previous method.

pkg/storage/kubernetes/ipam.go
```diff
@@ -2,31 +2,168 @@ package kubernetes
...
+
+	clientID := fmt.Sprintf("%s/%s", ipamConf.PodNamespace, ipamConf.PodName)
+
+	key := client.ObjectKey{
+		Name:      fmt.Sprintf("%s-%s", leasePrefix, leaseName),
+		Namespace: ipam.namespace,
+	}
```
maiqueb, line 143: Same here; you could re-use the helper method.

pkg/storage/kubernetes/ipam.go
```diff
@@ -2,31 +2,168 @@ package kubernetes
...
+	err := ipam.client.Get(ctx, key, lease)
+	if err != nil {
+		logging.Errorf("client %s, failed to get lease %s, err: %s", clientID, key.Name, err)
+		return err
+	}
+	// check if this client is the owner of the lease and update
+	if lease.Spec.HolderIdentity == nil || *lease.Spec.HolderIdentity != clientID {
+		return fmt.Errorf("client %s, lease %s already cleared or held by someone else",
+			clientID, lease.Name)
+	}
```
maiqueb, line 155: This would merit yet another helper - `func isLeaseOwnedBy(clientID string) bool {...}` .

pkg/storage/kubernetes/ipam.go
```diff
@@ -2,31 +2,168 @@ package kubernetes
...
+			logging.Errorf("failed to get lease: %s, error: %s", key.Name, err)
+			return nil, err
+		}
+		// create lease
+		lease = &coordinationv1.Lease{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      fmt.Sprintf("%s-%s", leasePrefix, leaseName),
+				Namespace: ipam.namespace},
+			Spec: coordinationv1.LeaseSpec{
+				LeaseDurationSeconds: pointer.Int32Ptr(leaseTimeout),
```
maiqueb, line 54: Is this really needed ?... Why don't you just `LeaseDurationSeconds: &leaseTimeout` ? 

Same for below; you can:
```golang
initialLeaseTransitions := uint32(0)
...
    Spec: coordinationv1.LeaseSpec{
        LeaseDurationSeconds: &leaseTimeout,
        LeaseTransitions:           &initialLeaseTransitions,
    },
...
```

I think this would remove the need for the `"k8s.io/utils/pointer"` dependency.

maiqueb (commented):
I am OK with the overall approach, and I think it might be a very helpful fix to improve the performance in the short term while we work through an architecture change for long term optimization.

I got some ideas on how to (IMHO) improve the readability of the code in the mean time, please consider them.

One thing I have to point out is that I miss testing for this change - not sure how easy it will be to unit test / or e2e test it.

martinkennelly:
@rthakur-est Awesome perf boost. I want to understand your perf numbers - did you use a range per cluster? I see you have a lease per range.

Also, I am wondering what if multiple instances go for the lease at the same time (because holdersidentity expired or no holderidentity). Is there a possibility multiple instances of whereabouts think they are the leases holderIdentity? Trying to understand the risk there. Last write in that race would be the holderidentity of the lease.

What happens if an instance of whereabouts gets the lease but their ownership of the lease has expired (some other instance now owns it) and they make an update to the database. Is there a prevention against this here (even if its very unlikely)?

Thanks again!

rthakur-est:
> I am OK with the overall approach, and I think it might be a very helpful fix to improve the performance in the short term while we work through an architecture change for long term optimization.
> 
> I got some ideas on how to (IMHO) improve the readability of the code in the mean time, please consider them.
> 
> One thing I have to point out is that I miss testing for this change - not sure how easy it will be to unit test / or e2e test it.

Thanks for providing your valuable comments. I will work on resolving those.
I am also not sure what test can be added for this.

rthakur-est:
> @rthakur-est Awesome perf boost. I want to understand your perf numbers - did you use a range per cluster? I see you have a lease per range.
The lease is per ip range.
> 
> Also, I am wondering what if multiple instances go for the lease at the same time (because holdersidentity expired or no holderidentity). Is there a possibility multiple instances of whereabouts think they are the leases holderIdentity? Trying to understand the risk there. Last write in that race would be the holderidentity of the lease.
> 
Once the lease expires, other instances will contend for the lease but only one instance will be able to sucessfully get it. Others will get conflict error.

> What happens if an instance of whereabouts gets the lease but their ownership of the lease has expired (some other instance now owns it) and they make an update to the database. Is there a prevention against this here (even if its very unlikely)?
> 
This scenario is quite possible and will result in duplicate IP assignment. I think for protection, the context timeout should not exceed the lease timeout. So, once lease expires and context timeout should also occur and it should prevent the write to the database.
> Thanks again!

rthakur-est:
> @rthakur-est: It's great that your lease lock approach succesfully prevents both duplicate and stale IPs. Could you, for reference and comparison, also provide measurements on the same 100 node setup for a) the original faulty SW baseline (prior to the leader election fix) and report the number of duplicates and stales and b) the etcd storage backend?

We tested the 1st scenario with v0.4.2 release on the 100 node setup.
Deployment took 1min 32s.
Undeployment took 1min 42s.
There were 280 duplicate IPs on deployment and 84 stale IPs on undeployment.

Couldn't test the 2nd scenario due to unavailability of the setup.

martinkennelly:
> The lease is per ip range.

Thank you. How many leases were used in the test?

> Once the lease expires, other instances will contend for the lease but only one instance will be able to sucessfully get it. Others will get conflict error.

Is this where that could occur in the [code](https://github.com/k8snetworkplumbingwg/whereabouts/pull/127/files#diff-7a77fc45b536e504372eba572994f95f478aa250d4cdb936bdc3d1018c86b09eR116)?

I am wondering what happens if two whereabouts instances perform this check at the same time:
[`if lease.Spec.HolderIdentity == nil || isLeaseExpired {`](https://github.com/k8snetworkplumbingwg/whereabouts/pull/127/files#diff-7a77fc45b536e504372eba572994f95f478aa250d4cdb936bdc3d1018c86b09eR107)

Both evaluate it to be true and proceed with acquiring the lease. Will there be an error on one of them when they call the update function that follows? I am worried, that the last instance that makes the [Update](https://github.com/k8snetworkplumbingwg/whereabouts/pull/127/files#diff-7a77fc45b536e504372eba572994f95f478aa250d4cdb936bdc3d1018c86b09eR112) call will acquire the lease and neither of the two instances get an error. 

_If_ this is true, then before whereabouts makes an update to the backend that it is reserving a particular ip, it must make sure it is the current lease holder and it must ensure it does this within a safe time window before the expiration of its ownership of the lease otherwise it should retry.

I hope, I am misunderstanding.. but this can be solved anyway by making a check before writing to backend.
Thanks again for you replies.

rthakur-est:
> > The lease is per ip range.
> 
> Thank you. How many leases were used in the test?
> 
We had one ip range. S, only one was used.

> > Once the lease expires, other instances will contend for the lease but only one instance will be able to sucessfully get it. Others will get conflict error.
> 
> Is this where that could occur in the [code](https://github.com/k8snetworkplumbingwg/whereabouts/pull/127/files#diff-7a77fc45b536e504372eba572994f95f478aa250d4cdb936bdc3d1018c86b09eR116)?
> 
> I am wondering what happens if two whereabouts instances perform this check at the same time:
> [`if lease.Spec.HolderIdentity == nil || isLeaseExpired {`](https://github.com/k8snetworkplumbingwg/whereabouts/pull/127/files#diff-7a77fc45b536e504372eba572994f95f478aa250d4cdb936bdc3d1018c86b09eR107)
> 
> Both evaluate it to be true and proceed with acquiring the lease. Will there be an error on one of them when they call the update function that follows? I am worried, that the last instance that makes the [Update](https://github.com/k8snetworkplumbingwg/whereabouts/pull/127/files#diff-7a77fc45b536e504372eba572994f95f478aa250d4cdb936bdc3d1018c86b09eR112) call will acquire the lease and neither of the two instances get an error.
> 
> _If_ this is true, then before whereabouts makes an update to the backend that it is reserving a particular ip, it must make sure it is the current lease holder and it must ensure it does this within a safe time window before the expiration of its ownership of the lease otherwise it should retry.
> 
> I hope, I am misunderstanding.. but this can be solved anyway by making a check before writing to backend.
Its entirely possible that two or more clients see that the lease is not owned by anyone and proceed to call update() to acquire it. However k8s optimistic concurrency control will ensure that one of them will succeed and rest will get conflict error. 

> Thanks again for you replies.

martinkennelly:
> However k8s optimistic concurrency control will ensure that one of them will succeed and rest will get conflict error.

@rthakur-est Makes sense now after looking it up - thank you!

pkg/storage/kubernetes/client.go
```diff
@@ -25,13 +26,16 @@ type Client struct {
...
+	coordinationv1.AddToScheme(scheme)
 
 	config, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
 		&clientcmd.ClientConfigLoadingRules{ExplicitPath: kubeconfigPath},
 		&clientcmd.ConfigOverrides{}).ClientConfig()
 	if err != nil {
 		return nil, err
 	}
+	config.QPS = 50.0
+	config.Burst = 100
```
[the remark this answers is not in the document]
rthakur-est: client-go does rate limiting of requests towards api server and default qps/burst are 5/10.
https://pkg.go.dev/k8s.io/client-go/rest?utm_source=godoc#Config
Since, plugin ends up doing lots of requests towards api server, I changed the values.

pkg/storage/kubernetes/ipam.go
```diff
@@ -2,31 +2,168 @@ package kubernetes
...
+)
+
+func getLease(ctx context.Context, ipam *KubernetesIPAM, leaseName string) (*coordinationv1.Lease, error) {
+	lease := &coordinationv1.Lease{}
+	c := ipam.client
+	leaseTimeout := int32(ipam.config.LeaseDuration)
+
+	key := client.ObjectKey{
+		Name:      fmt.Sprintf("%s-%s", leasePrefix, leaseName),
+		Namespace: ipam.namespace}
```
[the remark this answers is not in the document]
rthakur-est: Done

pkg/storage/kubernetes/ipam.go
```diff
@@ -2,31 +2,168 @@ package kubernetes
...
+			if !errors.IsAlreadyExists(err) {
+				logging.Errorf("create lease: %s, error: %s", key.Name, err)
+				return nil, err
+			}
+			// fetch the lease
+			err = c.Get(ctx, key, lease)
+			if err != nil {
+				logging.Errorf("get lease %s after creation, error: %s", key.Name, err)
+				return nil, err
+			}
```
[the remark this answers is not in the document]
rthakur-est: resourceVersion from the lease object is required.

pkg/storage/kubernetes/ipam.go
```diff
@@ -2,31 +2,168 @@ package kubernetes
...
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	leasePrefix = "whereabouts"
+)
+
+func getLease(ctx context.Context, ipam *KubernetesIPAM, leaseName string) (*coordinationv1.Lease, error) {
```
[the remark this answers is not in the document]
rthakur-est: I have made the lease functions as methods of the KubernetesIPAM struct.

pkg/storage/kubernetes/ipam.go
```diff
@@ -2,31 +2,168 @@ package kubernetes
...
+	return lease, nil
+}
+
+func acquireLease(ipam *KubernetesIPAM, ipamConf *whereaboutstypes.IPAMConfig) error {
+	var lease *coordinationv1.Lease
+	var err error
+
+	backoff := int64(ipamConf.Backoff)
+
+	clientID := fmt.Sprintf("%s/%s", ipamConf.PodNamespace, ipamConf.PodName)
```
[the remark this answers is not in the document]
rthakur-est: changed to holderIdentity based on the field in lease spec.

pkg/storage/kubernetes/ipam.go
```diff
@@ -2,31 +2,168 @@ package kubernetes
...
+		if lease.Spec.HolderIdentity != nil {
+			acquireTime := lease.Spec.AcquireTime
+			leaseDuration := *lease.Spec.LeaseDurationSeconds
+
+			if int32(time.Since(acquireTime.Time).Seconds()) >= leaseDuration {
+				logging.Debugf("lease %s held by %s expired, acquireTime: %v",
+					lease.Name, *lease.Spec.HolderIdentity, acquireTime)
+				isLeaseExpired = true
+			}
+		}
```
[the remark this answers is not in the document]
rthakur-est: Done.

pkg/storage/kubernetes/ipam.go
rthakur-est: Done

pkg/storage/kubernetes/ipam.go
rthakur-est: Done

pkg/storage/kubernetes/ipam.go
rthakur-est: Done

pkg/storage/kubernetes/ipam.go
rthakur-est: It was not required. I have made the suggested change.

rthakur-est:
@dougbtv can you please review this PR?

dougbtv:
I believe this is related to the work that was merged for the lease. Thanks for your contribution towards the final conclusion (if this is closed in error, please reopen with an explanation, thanks much!)

PR #118, opened 2021-07-05
IP ranges cleanup

maiqueb:
This PR adds a cronjob that runs every 5 minutes, and checks the consistency of the IP addresses currently assigned by the pools.

TODO:
  - retrieve a list of `Pods` once instead of checking N times if pods exist
  - ensure the `ip-reconciler` runs at most a little bit less that the provided cron expression
  - missing unit tests (I'd have to mock the datastore, and I didn't want to establish that precedent...) *but* it is tested e2e

PR #128, opened 2021-08-02
fix: Skip default network in ip reconciler

carezkh:


dougbtv:
LGTM. Going to get another set of eyes too. Thanks for the contribution!

pkg/reconciler/wrappedPod.go
```diff
@@ -67,6 +67,10 @@ func getFlatIPSet(pod v1.Pod) map[string]void {
 
 	for _, network := range networkStatusList {
 		// we're only after multus secondary interfaces
+		if len(network.Interface) == 0 {
```
maiqueb, line 70: hm, right, never had considered a network without interfaces. Can that happen ?..

This code positioned here renders the above comment invalid; afaiu, all multus secondary interfaces will have the `net` prefix followed by an ordinal.

pkg/reconciler/wrappedPod.go
```diff
@@ -67,6 +67,10 @@ func getFlatIPSet(pod v1.Pod) map[string]void {
...
+			continue
+		}
+		// TODO: get whereabouts interface names from pod annotations
```
maiqueb, line 73: I don't understand this comment.

maiqueb (changes requested):

pkg/reconciler/wrappedPod.go
maiqueb: can you add a test in https://github.com/k8snetworkplumbingwg/whereabouts/blob/master/pkg/reconciler/wrappedPod_test.go#L17 that triggers this fix ?

carezkh: In my case, the default pod interface will be displayed in network status annotations without field interface be filled.
like below,
`    k8s.v1.cni.cncf.io/network-status: |-
      [{
          "name": "",
          "ips": [
              "172.16.125.147"
          ],
          "default": true,
          "dns": {}
      },{
          "name": "default/bridge-whereabouts",
          "interface": "net0",
          "ips": [
              "192.168.122.101"
          ],
          "mac": "ea:65:54:2e:a6:e3",
          "dns": {}
...`

carezkh: The multus secondary interfaces can be named by user.

> apiVersion: v1
kind: Pod
metadata:
  name: samplepod
  annotations:
    k8s.v1.cni.cncf.io/networks: bridge-whereabouts@hello,bridge-hostlocal@world
...

pkg/reconciler/wrappedPod.go
carezkh: The multus secondary interfaces can be named by user. Maybe the interface names should be get from networks annotations. Otherwise, for custome interface name 'hello', ip reconciler will cleanup the ip and assign it to other pod.

BTW, a judgment for whether the interface is served by whereabout could be better.

pkg/reconciler/wrappedPod.go
maiqueb: I mean the interface names on the pods, that will appear listed on the network status, not the networks.

pkg/reconciler/wrappedPod.go
maiqueb: I still do not understand how we can have an interface within the pod named other than `netX` . Those are the names that will be reflected on the network status annotations.

pkg/reconciler/wrappedPod.go
carezkh: the interface `eth0` (not multus secondary interface) with ip `172.16.125.147` will appear listed on the network status like below
`k8s.v1.cni.cncf.io/network-status: |- [{ "name": "", "ips": [ "172.16.125.147" ], "default": true, "dns": {} },....`

pkg/reconciler/wrappedPod.go
carezkh: https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/how-to-use.md#launch-pod-with-text-annotation-with-interface-name

pkg/reconciler/wrappedPod.go
maiqueb: The reconciler **only** tackles multus secondary interfaces, as per https://github.com/k8snetworkplumbingwg/whereabouts/pull/118#discussion_r670518554. 

Reconciling the primary interface was never a goal.

How about you skip the interface on the network status when it is the `default` network, rather than when its interface does not have a name ? I find that more explicit.

I'm still missing a unit test for this fix, and I wonder if we should not add a log instead of silently ignoring the unnamed interfaces.

pkg/reconciler/wrappedPod.go
maiqueb: I see. I was not aware of this, thanks for taking the time to lookup this document.

I'd rather you post an issue than adding a todo in the code.

pkg/reconciler/wrappedPod.go
carezkh: Right, thanks for your advice! And i'll add a unit test for this fix.

How about just ignoring it? Does the log info here have any help?

pkg/reconciler/wrappedPod.go
carezkh: Thanks for your advice, i'll post an issue and remove the todo in the code.

pkg/reconciler/wrappedPod.go
maiqueb: I'm unsure - which I guess makes your suggestion as good / bad as mine :) 

We can silently ignore it ... It's not an error.

maiqueb (approved):
Thank you!

The test looks really good :)

dougbtv:
Thanks Miguel and @carezkh for the contribution :+1:

PR #129, opened 2021-08-03
Add whereabouts-cleaner.sh script

TothFerenc:
Hi,

Hereby I share my Whereabouts cleaner implementation. I know that `reconciler` command is being developed to address the stale IP reservations (first feature below). But this script handles the IP duplication issue as well (second feature below). Maybe this script is useful until the go implementation (`reconciler`) covers all the problematic cases we have in field deployments.

Features:
- It detects if a reserved IP is not used by the referenced Pod anymore. In this case the IP is released. Special StatefulSet Pod case is handled (same Pod name with different IP).
- It detects if Whereabouts served the same IP to multiple Pods (due to missing K8s locking implementation). It deletes all the non-referenced Pods using the same IP address. The restarted Pods will get new IPs from the pool.

Usage: Let it run continuously in an un-attended terminal (ideally in a Pod dedicated for this purpose).

Pre-requisites:
- Recent Whereabouts version, which contains the #86 patch
- kubectl with permissions to query/manage Pods and IPPool CRD
- Python 2/3 with `ipaddress` module
- `bash` & `jq`


maiqueb:
Does this PR still make sense, now that the reconciler can be deployed as a cronjob ? 

Do you find anything missing from it ? If so, we can collaborate and add it there.

TothFerenc:
> While we're currently deploying the reconciler on a `cronjob` - that indeed does have the limitation you meantion - there is an ongoing discussion about introducing a controller for whereabouts. If that is accepted, the reconciler functionality will eventually be moved there.

But that is not ready at the moment, while the latest Whereabouts release can serve duplicate IPs, so I need a quick solution now to be able to operate big clusters.

> Plus, I also see negative impacts in running the reconcile loop too often in large clusters - actually, even once every minute seems too much to me.

What are those negative impacts? Can you please elaborate more?
I prefer to run frequent reconcile cycles to be able to fix the problems during a helm chart deployment, so helm won't timeout due to bad Pod states (caused by duplicate IPs or exhausted pools/ranges).
 
> Lastly, do you think you can reconcile all the whereabouts IPs on a large cluster under 10 seconds ? IMO, that is too ambitious, and puts a lot of pressure the kubernetes API.

My implementation does not contain any parallelization/threading. Once a reconcile task is done (taking the time as much as it is needed to complete), there's a 10 sec sleep before the next reconcile task is started.
In large clusters we have dedicated master nodes, their cpu consumption is not an issue.

dougbtv:
This is replaced by the reconciler / reconciliation control loop.

PR #145, opened 2021-09-20
Release lease lock when LE ends

martinkennelly:
Signed-off-by: Martin Kennelly <mkennell@redhat.com>

Description of feature: https://pkg.go.dev/k8s.io/client-go/tools/leaderelection#LeaderElectionConfig

Enabling this should be ok because we are not interacting with our protected resource (IP Pools) when we end LE.

### DIFF
--- a/pkg/storage/kubernetes/ipam.go
+++ b/pkg/storage/kubernetes/ipam.go
@@ -313,6 +313,7 @@ func newLeaderElector(clientset *kubernetes.Clientset, namespace string, podName
 		LeaseDuration: time.Duration(leaseDuration) * time.Millisecond,
 		RenewDeadline: time.Duration(renewDeadline) * time.Millisecond,
 		RetryPeriod:   time.Duration(retryPeriod) * time.Millisecond,
+		ReleaseOnCancel: true,
 		Callbacks: leaderelection.LeaderCallbacks{
 			OnStartedLeading: func(_ context.Context) {
 				logging.Debugf("OnStartedLeading() called")


s1061123 (approved):
LGTM, thanks!

pkg/storage/kubernetes/ipam.go
```diff
@@ -313,6 +313,7 @@ func newLeaderElector(clientset *kubernetes.Clientset, namespace string, podName
 		LeaseDuration: time.Duration(leaseDuration) * time.Millisecond,
 		RenewDeadline: time.Duration(renewDeadline) * time.Millisecond,
 		RetryPeriod:   time.Duration(retryPeriod) * time.Millisecond,
+		ReleaseOnCancel: true,
```
maiqueb, line 316: I think the commit msg should explicitly quote the name of the attribute - `ReleaseOnCancel`.

Furthermore, while I think this is an amazing find, and in the harmless, I see in the [LE docs](https://github.com/kubernetes/client-go/blob/v0.22.2/tools/leaderelection/leaderelection.go#L150)  that `If you set this to true, you must ensure all code guarded by this lease has successfully completed prior to cancelling the context` . 

I get a bit confused by their wording, especially with "successfully completed." Can we guarantee that ?..

rthakur-est: I agree with @maiqueb, the leader lease should be held till the update to ippools CRD is completed. If cancellation happens due to timeout before update is complete, duplicate IP allocation might occur.

martinkennelly: > I get a bit confused by their wording, especially with "successfully completed." Can we guarantee that ?..
I believe we can. The LE context being "canceled" for ```le.Run()``` should only happen when case <-leader is guaranteed to have completed or we can be certain its not the case being triggered. Current master, the LE context is only called when case <- leader is finished so we are certain it has completed, therefore its good to go into master now.
Any other PRs (like #142 ) need, we need to keep this in mind, and I think I have.

martinkennelly:
@maiqueb forced pushed to update the commit message. LMK if its ok.

dougbtv (approved):
LGTM!

e0ne (approved):

Issue #130, opened 2021-08-04
IP will conflict when the name of multus secondary interface without prefix 'net'


carezkh:
Hi everybody! 

When i name a multus secondary interface without prefix `net`, for example `hello`. Whereabouts will give `hello` an ip, for example `192.168.122.100`.

This ip will be cleaned up from ippools by ip reconciler, and it will be assigned to other pod, which causes ip conflict!

>  2021-08-04T11:15:04Z [debug] NewReconcileLooper - Kubernetes config file located at: /host/etc/cni/net.d/whereabouts.d/whereabouts
│ 2021-08-04T11:15:06Z [debug] successfully read the kubernetes configuration file located at: /host/etc/cni/net.d/whereabouts.d/whe
│ 2021-08-04T11:15:06Z [debug] listing IP pools
│ 2021-08-04T11:15:06Z [debug] the IP reservation: IP: 192.168.122.100 is reserved for pod: default/samplepod
│ 2021-08-04T11:15:06Z [debug] pod reference default/samplepod matches allocation; Allocation IP: 192.168.122.100; PodIPs: map[]
│ 2021-08-04T11:15:06Z [debug] pod ref default/samplepod is not listed in the live pods list
│ 2021-08-04T11:15:06Z [debug] pod reference default/samplepod matches allocation; Allocation IP: 192.168.122.100; PodIPs: map[]
│ 2021-08-04T11:15:06Z [debug] pod ref default/samplepod is not listed in the live pods list
│ 2021-08-04T11:15:06Z [debug] Going to update the reserve list to: []
│ 2021-08-04T11:15:06Z [debug] successfully cleanup IPs: [192.168.122.100]
│ 2021-08-04T11:15:07Z [debug] removed stale overlappingIP allocation [192.168.122.100]

The ip reconciler will only get ip set from interface with prefix `net`, which causes this bug!
See https://github.com/k8snetworkplumbingwg/whereabouts/blob/master/pkg/reconciler/wrappedPod.go#L68

xagent003:
+1 I am seeing this too.

It does not happen when using the default interface name of "net1". But in the Pod spec, if we override the interface name to something like "eth1", 5 minutes later after the ip-reconciler has run, I see the IP Pool allocation has been cleaned up.

[...]

PR #147, opened 2021-09-21
ip-reconciler: Add all non default interfaces to Pod IP list

xagent003:
This fixes https://github.com/k8snetworkplumbingwg/whereabouts/issues/130

In code below,

```
	for _, network := range networkStatusList {
		// we're only after multus secondary interfaces
		if network.Default {
			continue
		}

		if network.Interface[:multusPrefixSize] == multusInterfaceNamePrefix {
			for _, ip := range network.IPs {
				ipSet[ip] = empty
			}
		}
	}
```

I don't know of any case where there would be a network attachment that isn't default, but not a Multus attachment? The way I understand it, we have a primary CNI (such as Calico, for us). This is implicitly attached to every Pod as "eth0", and in describe Pod, it is always set to default: True

So, is there any case where the Pod would have multiple Networks, and one of them not be Default yet not Multus?

Either way, I don't think this check is correct, as in the Pod's network annotation, it is allowed to change the interface name. At very least there should be some other way to detect Multus interfaces (if simply checking for not default isn't sufficient). For example, see this example: https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/how-to-use.md#launch-pod-with-json-annotation-with-interface.

With my fix, I set Pod's interface to be named "eth1" for my macvlan network. Per the Log I added to ip-reconciler, I can see it's now added the the IPSet for the Pod:

```
xagent@pop-os:~/luigi/samples/macvlan$ kubectl logs ip-reconciler-1632195360-9vgrk -n kube-system
2021-09-21T03:36:07Z [debug] NewReconcileLooper - Kubernetes config file located at: /host/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig
2021-09-21T03:36:08Z [debug] successfully read the kubernetes configuration file located at: /host/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig
2021-09-21T03:36:08Z [debug] listing IP pools
2021-09-21T03:36:08Z [debug] **Added IP 10.128.165.230 for pod kube-system/samplepod1**
2021-09-21T03:36:08Z [debug] **Added IP 10.128.165.231 for pod kube-system/samplepod2**
2021-09-21T03:36:08Z [debug] the IP reservation: IP: 10.128.165.230 is reserved for pod: kube-system/samplepod1
2021-09-21T03:36:08Z [debug] pod reference kube-system/samplepod1 matches allocation; Allocation IP: 10.128.165.230; PodIPs: map[10.128.165.230:{}]
2021-09-21T03:36:08Z [debug] the IP reservation: IP: 10.128.165.231 is reserved for pod: kube-system/samplepod2
2021-09-21T03:36:08Z [debug] pod reference kube-system/samplepod2 matches allocation; Allocation IP: 10.128.165.231; PodIPs: map[10.128.165.231:{}]
2021-09-21T03:36:09Z [debug] pod reference kube-system/samplepod1 matches allocation; Allocation IP: 10.128.165.230; PodIPs: map[10.128.165.230:{}]
2021-09-21T03:36:09Z [debug] pod reference kube-system/samplepod2 matches allocation; Allocation IP: 10.128.165.231; PodIPs: map[10.128.165.231:{}]
2021-09-21T03:36:09Z [debug] no IP addresses to cleanup

```

I just removed the entire testcase for this too, as it's called "should filter out non-multus annotations", but to generate it, it calls function "generateMultusNetworkStatus". It just sets a different interface name as "eth0", which is allowed by Multus.

### DIFF
--- a/pkg/reconciler/wrappedPod.go
+++ b/pkg/reconciler/wrappedPod.go
@@ -71,10 +71,9 @@ func getFlatIPSet(pod v1.Pod) map[string]void {
 			continue
 		}
 
-		if network.Interface[:multusPrefixSize] == multusInterfaceNamePrefix {
-			for _, ip := range network.IPs {
-				ipSet[ip] = empty
-			}
+		for _, ip := range network.IPs {
+			ipSet[ip] = empty
+			logging.Debugf("Added IP %s for pod %s", ip, composePodRef(pod))
 		}
 	}
 	return ipSet
--- a/pkg/reconciler/wrappedPod_test.go
+++ b/pkg/reconciler/wrappedPod_test.go
@@ -114,23 +114,6 @@ var _ = Describe("Pod Wrapper operations", func() {
 			Expect(podSecondaryIPs).To(Equal(map[string]void{"192.168.14.14": {}, "10.10.10.10": {}}))
 		})
 
-		It("should filter out non-multus annotations", func() {
-			secondaryIfacesNetworkStatuses := generateMultusNetworkStatusList("192.168.14.14", "10.10.10.10")
-
-			networkStatus := append(
-				secondaryIfacesNetworkStatuses,
-				generateMultusNetworkStatus("eth0", "network33", "14.15.16.20"))
-			pod := v1.Pod{
-				ObjectMeta: metav1.ObjectMeta{
-					Annotations: generateMultusNetworkStatusAnnotationFromNetworkStatus(networkStatus...),
-				},
-			}
-
-			podSecondaryIPs := wrapPod(pod).ips
-			Expect(podSecondaryIPs).To(HaveLen(2))
-			Expect(podSecondaryIPs).To(Equal(map[string]void{"192.168.14.14": {}, "10.10.10.10": {}}))
-		})
-
 		It("return an empty list when the network annotations of a pod are invalid", func() {
 			pod := v1.Pod{
 				ObjectMeta: metav1.ObjectMeta{


pkg/reconciler/wrappedPod.go
```diff
@@ -71,10 +71,9 @@ func getFlatIPSet(pod v1.Pod) map[string]void {
 			continue
 		}
 
-		if network.Interface[:multusPrefixSize] == multusInterfaceNamePrefix {
-			for _, ip := range network.IPs {
-				ipSet[ip] = empty
-			}
+		for _, ip := range network.IPs {
```
carezkh, line 74: What if the secondary interface is not managed by IPAM CNI whereabouts? Such as dhcp, host-local...

maiqueb: This is a valid concern, but, I think it is not introduced by this PR. 

Furthermore, iiuc, the scenario you point out @carezkh would only trigger something nasty (the reconciler would think IP A is still in use, when in truth it is was served by another IPAM plugin) **if** the user is using the same subnet in multiple networks, something which I think cannot be a use case. Am I right ? If so, we could probably get away with documenting this "limitation".

maiqueb (commented):
Thank you for this PR.

While I do not think this PR is wrong, I want to understand better @carezkh 's concerns.

pkg/reconciler/wrappedPod.go
carezkh: In my opinion, user can use the same subnet in multiple networks when **the underlay networks of multiple NetworkAttachmentDefinitions are isolated** (maybe different storage networks or VLANs).

Example like below,
`cat macvlan-whereabouts1.yaml`

> apiVersion: k8s.cni.cncf.io/v1
metadata:
  name: macvlan-whereabouts1
spec:
  config: '{
      "cniVersion": "0.3.1",
      "type": "macvlan",
      "master": "eth0",
      "mode": "bridge",
      "ipam": {
          "type": "whereabouts",
          "range": "192.168.0.100/24",
          "gateway": "192.168.0.1"
      }
    }'

`cat macvlan-whereabouts2.yaml`

> apiVersion: k8s.cni.cncf.io/v1
metadata:
  name: macvlan-whereabouts2
spec:
  config: '{
      "cniVersion": "0.3.1",
      "type": "macvlan",
      "master": "eth1",
      "mode": "bridge",
      "ipam": {
          "type": "whereabouts",
          "range": "192.168.0.100/24",
          "gateway": "192.168.0.1"
      }
    }'

If the underlay networks of interface `eth0` and `eth1` are different storage network, the same subnet is ok. Am i rignt?

However, it's a little hard to tell if the network is isolated, maybe globally unique IP address is ok.

maiqueb: Given that we (at least I) have no clue as to how to detect (via any sort of api) if the underlying interfaces are connected, and the mess the routing tables would like in the pods, I personally think the safest course of action is to assume globally unique IP addresses and document that.

I don't think this breaks the CNI spec, but I'm unsure about the de-facto standard for Kubernetes Network Custom Resource Definition.

carezkh: If the mess of the routing tables (in pods) caused by imporper use is inevitable, i think it should be handled by net adminitrator instead of CNI plugin.

Net admin plans and deploys the network and instructs others in its proper use, and IPAM CNI can assign IP addresses from pool without caring about globally IP overlapping. 

Anyway, the globally unique IP addresses (what this PR has done) is ok.

xagent003: > What if the secondary interface is not managed by IPAM CNI whereabouts? Such as dhcp, host-local...

I think that's true, but that issue would still exist without this fix. The "net" naming is coming from multus and not IPAM. So without this fix, the ip-reconciler is broken and a blocker as it automatically cleans up any secondary Multus interface not named "net". Which results in whereabouts assigning duplicate IPs and IP pool never filling up.

I think the point you raised is a separate issue with ip-reconciler

carezkh: I agree.

dougbtv (approved):
LGTM! Thanks for the patience.

PR #148, opened 2021-09-21
Add E2E test cases and execute with GH workflow

martinkennelly:
Test cases are simple and check that:
* IPs are unique
* Number of IP pool allocations reflect number of Pod IPs
allocated
* No stale allocations in IP pool following testing

Signed-off-by: Martin Kennelly <mkennell@redhat.com>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,3 +59,26 @@ jobs:
       - uses: shogo82148/actions-goveralls@v1
         with:
           parallel-finished: true
+
+  e2e-test:
+    name: e2e test
+    runs-on: ubuntu-latest
+    env:
+      NUMBER_OF_COMPUTE_NODES: 5
+    steps:
+      - name: Set up Go version
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.16
+
+      - name: Checkout code into the Go module directory
+        uses: actions/checkout@v2
+
+      - name: Install requirements
+        run: sudo apt-get install nmap jq
+
+      - name: Get tools, setup KinD cluster test environment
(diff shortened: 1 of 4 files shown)


maiqueb (commented):
While this addresses a huge gap in whereabouts, I don't think bash is a good choice for implementing the e2e tests.

Do we agree it would be preferable to have it on golang ? 

I don't having this in as is, and port these test cases to a proper programming language afterwards.

martinkennelly:
> Do we agree it would be preferable to have it on golang ?

100%

yeah - you're totally right it would be so much nicer in golang. I wanted to sleep at night with the changes I am currently making in the other PRs, so I setup this.

I think the setting up in the infra with KinD is nice to keep and would be used by golang test implementation. I believe @nicklesimba  is working on golang tests and can use that and discard the test shell scripts here.

maiqueb:
> > Do we agree it would be preferable to have it on golang ?
> 
> 100%
> 
> yeah - you're totally right it would be so much nicer in golang. I wanted to sleep at night with the changes I am currently making in the other PRs, so I setup this.
> 
> I think the setting up in the infra with KinD is nice to keep and would be used by golang test implementation. 

Absolutely.

> I believe @nicklesimba is working on golang tests and can use that and discard the test shell scripts here.

Didn't know; nice. I can collaborate on that as well, if needed.

dougbtv:
I'm OK with proceeding with having these as a "version 1", and then we'll come back as a second pass, and have @nicklesimba work through converting into golang-based tests. 

I think that while the bash tests aren't ideal, they are pragmatic to get some tests in place for the short run. That being said, I'm OK to approve and then follow-up, when we come out of WIP

dougbtv:
Giving this a CI re-run after merging #142 :+1:

martinkennelly:
Forced pushed to change multus deployment path.

PR #151, opened 2021-09-27
added sleep-for-race option for IPAM

nicklesimba:
Pods can be given IPAM field sleep_for_race, which if set to "true" will cause a pod to sleep for 20 seconds before updating IP pool. If another pod is spun up within those 20 seconds without sleep_for_race set to true, it should take the same IP address given to the first pod, and can serve as a test for whether race conditions are present.


dougbtv (approved):
LGTM!

PR #152, opened 2021-10-04
Fix test to include error when IP pool total doesn't equal assigned WB pod IPs

martinkennelly:
The tests check if IP pool IPs are seen in assigned WB pod IPs.
The tests check if pod IPs are seen in IP pool
but there maybe duplicate IP(s) assigned that isn't accounted for so we check if the total IPs assigned is equal to the total seen in IP pool.


dougbtv (approved):
LGTM! Thanks Martin.

PR #154, opened 2021-10-06
build: enable e2e testing for podman

maiqueb:
As per [0] kind `load docker-image` commands are docker specific.
This commit changes that instruction, using the generic
`load image-archive` instead, which works both for podman and docker.

[0] - https://github.com/kubernetes-sigs/kind/issues/2027


maiqueb:
@martinkennelly / @dougbtv 

Making podman users happy.

PR #156, opened 2021-10-06
e2e tests: fix currently broken setup on kind clusters

maiqueb:


maiqueb:
Just noticed the `disableCheck` attribute is only featured when a config list is provided: https://github.com/containernetworking/cni/blob/master/pkg/types/types.go#L82

PR #161, opened 2021-11-08
Configurable sleep_for_race parameter

nicklesimba:
Changed sleep_for_race field for IPAM. It is now an integer that will not affect a pod's launch if set to 0 (default), but will invoke sleep for n seconds based on what it is otherwise set to.


dougbtv (approved):
LGTM! Thanks for the update on this one.

PR #164, opened 2021-11-10
Fix configurability of OverlappingRanges

### LINKED ISSUE #163: Unable to disable OverlappingRanges using the config file
OverlappingRanges feature is enabled by default, this way: 
https://github.com/k8snetworkplumbingwg/whereabouts/blob/master/pkg/config/config.go#L38

Later on, the config file (flatfile) is processed and those values are merged to the actual IPAM config struct.
Once this OverlappingRanges is set to True, we cannot override it to False. I believe the problem is that a boolean variable has no "undefined" value, just True or False. In case of False, we don't know whether it is unset or intentionally set to False.

So we cannot disable the OverlappingRanges via config file, we need to disable it in every NAD, which is not a feasible option.

TothFerenc:
Boolean values are not handled properly during merging structs.
Better to Unmarshal the JSON input (again) on top of the existing
IPAM config (originating from the specified config file) to override
the provided values only.

Fixes #163

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -10,7 +10,6 @@ import (
 
 	cnitypes "github.com/containernetworking/cni/pkg/types"
 	types020 "github.com/containernetworking/cni/pkg/types/020"
-	"github.com/imdario/mergo"
 	"github.com/k8snetworkplumbingwg/whereabouts/pkg/logging"
 	"github.com/k8snetworkplumbingwg/whereabouts/pkg/types"
 )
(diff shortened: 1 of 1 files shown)


dougbtv (approved):
LGTM! Thanks Toth.

TothFerenc:
@dougbtv 
Are you waiting for others to review? Or just forgot to merge?
Thanks!

dougbtv:
This... did indeed get missed, could you rebase? Thanks!

PR #167, opened 2021-11-25
Safer reconciler spec

maiqueb:
This PR updates the reconciler spec:
  - forbid concurrent executions
  - remove failed/complete pods
  - do not retry on error

The reconciler is fairly API hungry; having it retry on errors, and allow more of them when the system fails just clogs the cluster further.

### DIFF
--- a/doc/crds/ip-reconciler-job.yaml
+++ b/doc/crds/ip-reconciler-job.yaml
@@ -7,6 +7,7 @@ metadata:
     tier: node
     app: whereabouts
 spec:
+  concurrencyPolicy: Forbid
   schedule: "*/5 * * * *"
   jobTemplate:
     spec:
--- a/doc/crds/ip-reconciler-job.yaml
+++ b/doc/crds/ip-reconciler-job.yaml
@@ -8,6 +8,7 @@ metadata:
     app: whereabouts
 spec:
   concurrencyPolicy: Forbid
+  successfulJobsHistoryLimit: 0
   schedule: "*/5 * * * *"
   jobTemplate:
     spec:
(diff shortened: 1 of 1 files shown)


doc/crds/ip-reconciler-job.yaml
```diff
@@ -7,9 +7,13 @@ metadata:
     tier: node
     app: whereabouts
 spec:
+  concurrencyPolicy: Forbid
+  failedJobsHistoryLimit: 0
```
maiqueb, line 11: not exactly sure we want to delete the failed pods. What if we want to debug them ?

dougbtv: I like the idea of maybe setting this to `1`? That way we can see one failure so we can debug it?

maiqueb: That's OK, but afaiu, 1 is the default. I can just remove the entry altogether.

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#cronjobspec-v1beta1-batch

dougbtv (approved):
LGTM! Thanks

xagent003:
@maiqueb @dougbtv 

I had question as to why successfulJobsHistoryLimit: 0 was added in the commit.

A Job can succeed, but still produce bugs, right? For example the two issues that I had caught where custom Multus interface names were cleaned up, as well as the duplicate IP issue (Don't reconcile Pods in pending phase). In those cases the Job completed without error. If it was cleaned up without error, there would be no way to debug it

maiqueb:
> @maiqueb @dougbtv
> 
> I had question as to why successfulJobsHistoryLimit: 0 was added in the commit.
> 
> A Job can succeed, but still produce bugs, right? For example the two issues that I had caught where custom Multus interface names were cleaned up, as well as the duplicate IP issue (Don't reconcile Pods in pending phase). In those cases the Job completed without error. If it was cleaned up, there would be no way to debug it.

Sorry for having missed this comment. You're welcome to send a patch to fix it @xagent003 .

PR #177, opened 2022-01-19
feat: Static ip allocations only

### LINKED ISSUE #111: manually assign macvlan IP to pod
Hi,
I'd like to manually assign macvlan IP to a pod. it is possible but the pod will have two macvlan IP, one is automatically be attached, another is from pod annotation, so it will waste ip resources.....
Is it possible to assign only one macvlan IP(which defined by me) to a pod?
here is my configuration:
```
# NetworkAttachmentDefinition
metadata:
  name: whereabouts-conf
spec:
  config: '{
      "cniVersion": "0.3.0",
      "name": "whereaboutsexample",
      "type": "macvlan",
      "master": "ens3",
      "mode": "bridge",
      "log_file": "/var/log/whereabouts.log",
      "log_level": "debug",
      "ipam": {
        "type": "whereabouts",
        "range": "172.32.3.20-172.32.3.29/24",
        "gateway": "172.32.3.1",
        "exclude": [
          "172.32.3.22/32"
        ],
        "routes": [
          { "dst": "0.0.0.0/0" }
        ],
        "gateway": "172.32.3.1"
      }
    }'

# pod definition
apiVersion: v1
kind: Pod
metadata:
  name: pod-1
  annotations:
    k8s.v1.cni.cncf.io/networks: '[{
      "name": "whereabouts-conf",
      "ips": [ "172.32.3.29/24" ],
      "mac": "c2:b0:57:aa:aa:aa",
      "gateway": [ "172.32.3.22" ]
    }]'
spec:
  containers:
  - name: pod-1
    command: ["/bin/ash", "-c", "trap : TERM INT; sleep infinity & wait"]
    image: 172.32.3.166/library/alpine
```

salkin:
When user wants only static IP address whereabouts should not reserve double addresses but put the static allocation into the list of reserved IPs.

Fixes #111

### DIFF
--- a/cmd/whereabouts.go
+++ b/cmd/whereabouts.go
@@ -48,7 +48,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 	result.Routes = ipamConf.Routes
 
 	logging.Debugf("Beginning IPAM for ContainerID: %v", args.ContainerID)
-	var newip net.IPNet
+	var newip []net.IPNet
 
 	ctx, cancel := context.WithTimeout(context.Background(), types.AddTimeLimit)
 	defer cancel()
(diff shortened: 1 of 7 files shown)


dougbtv:
I think the thing here that's... interesting is...

...the actual ability to allocate the static IP address is.... sort of an incidental feature, I didn't actually realize that worked, and that it carried over from some of the whereabouts origins (which started as a fork of the static ipam plugin)

So, I'm a little hesitant to bake it in as a feature.

That being said, I'd love to hear more about the usecase.

I think something that might be more fitting to the philosophy that we're using is an often requested feature which is for dual stack assignment, both assigning an IPv4 and an IPv6 address. This feature makes me think about having n-number-of-assigned-ranges as well. Which might fit the use case as well here?

You could assign in an /32 (for ipv4).

salkin:
The use case we have is when we integrate Telco VNFs with macvlan to real routing domains. The VNFs have IPs from a campus network that are given to us, we cannot choose freely. Those we set as static IPs for the pods. We cannot tell to our end users that "we by the way take two routable IPs from you". Therefore this is a showstopper for us to start using whereabouts. 

We have used the host-local CNI upto now but that of course has it's own challenges when those IPs are just stored in the local node. 

If the n-number-of-assigned ranges would support the same as this pull request that setting a static IP only reserves a single IP I am all in for it.

dougbtv:
I think we'll have the kind of flexibility to have multiple ranges in this: https://github.com/k8snetworkplumbingwg/whereabouts/pull/237

If you don't think this will handle your use case, could you submit a design proposal such as #237? Thanks!

Issue #172, opened 2022-01-06
ip-reconciler fails with "context deadline exceeded" while listing IPPools


xagent003:
The ip-reconciler is repeatedly failing (not just a one off, rare Job). Pod shows in CrashBackoffLoop state, and when we inspect the container logs:

```
2022-01-05T07:25:24Z [debug] NewReconcileLooper - Kubernetes config file located at: /host/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig

2022-01-05T07:25:25Z [debug] successfully read the kubernetes configuration file located at: /host/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig

2022-01-05T07:25:34Z [debug] listing IP pools

2022-01-05T07:25:35Z [error] failed to retrieve all IP pools: context deadline exceeded

2022-01-05T07:25:35Z [error] failed to create the reconcile looper: failed to retrieve all IP pools: context deadline exceeded
```

API server is up and no other workloads show this type of error. The resources are fetch'able via `kubectl get ippools.whereabouts.cni.cncf.io` - to list all pools, or a specific one in detail. 

Does the context timeout need to be reset, is the initial timeout value too small? Should we pass in a dummy context.TODO() before making the client call to get IPPool?

xagent003:
The ListPods may be taking a long time, 9 seconds in this case. Should we increase the timeout or should we be generating a new context for each client API call that takes in a context?

@maiqueb ?

maiqueb:
> The ListPods may be taking a long time, 9 seconds in this case. Should we increase the timeout or should we be generating a new context for each client API call that takes in a context?
> 
> @maiqueb ?

I was sloppy enough when I wrote the to use `context.TODO` when listing the pods. I honestly think that is the single reason why we haven't seen this issue before - and at a larger scale.

Would you check #186 ? Odds are it fixes this issue.

PR #179, opened 2022-01-20
Pass in context.TODO while listing IPPools

xagent003:
Fixes Issue #172 : https://github.com/k8snetworkplumbingwg/whereabouts/issues/172

### DIFF
--- a/pkg/storage/kubernetes/client.go
+++ b/pkg/storage/kubernetes/client.go
@@ -80,7 +80,7 @@ func (i *Client) ListIPPools(ctx context.Context) ([]storage.IPPool, error) {
 	logging.Debugf("listing IP pools")
 	ipPoolList := &whereaboutsv1alpha1.IPPoolList{}
 
-	if err := i.client.List(ctx, ipPoolList, &client.ListOptions{}); err != nil {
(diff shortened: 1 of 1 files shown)


maiqueb:
> Fixes Issue #172 : #172

@xagent003 Is this PR working ? Does using a new context per request fix issue #172 ?

maiqueb:
While thinking about the underlying issue, it seems to me that we should still have a cancel-able context that respects a dead-line, something which, afaiu, `TODO` is not: https://github.com/golang/go/blob/master/src/context/context.go#L200

My preference would be for the timeout value - which is now hard-coded to 10 seconds - to be configurable (overridden maybe?) via the reconciler command line arguments.

Furthermore, re-using the already configured context does not seem like a good idea now. We should probably have a stripped down top-level parent context, and use new cancel-able with timeout for the API queries.

Your thoughts @xagent003 ? 

Before I forget: huge thanks for the issues / discussions / fixes @xagent003 .

xagent003:
Is this still valid, or should I close it out, @maiqueb ?

maiqueb:
> Is this still valid, or should I close it out, @maiqueb ?

I think you can close it since #186 and #209 were merged.

maiqueb:
Closing in lieu of #186 and #209 ( both already merged).

PR #181, opened 2022-01-25
Whereabouts comprehensive go e2e test

nicklesimba:
This PR introduces a more complex test that is based off of Martin's bash test - but is written in Golang. Once merged, Martin's test should be removed and replaced by this. The goal here is that a Golang based e2e will become more easily maintainable and readable in the long term.


go.mod
```diff
@@ -4,21 +4,38 @@ go 1.15
...
+	golang.org/x/tools v0.1.8 // indirect
 	gomodules.xyz/jsonpatch/v2 v2.1.0
-	k8s.io/api v0.20.2
-	k8s.io/apimachinery v0.20.2
-	k8s.io/client-go v0.20.2
+	google.golang.org/appengine v1.6.7 // indirect
+	honnef.co/go/tools v0.0.1-2020.1.4 // indirect
+	k8s.io/api v0.23.1
+	k8s.io/apimachinery v0.23.1
+	k8s.io/client-go v0.23.0-alpha.0
```
maiqueb, line 31: I'm concerned about this; IIUC, this would make us only accept k8s-1.23 API, which is the latest one. 

K8s supports the last 3 releases at all times - right now, it would be 1.23, 1.22, 1.21.

I don't think this bump can happen at this moment.

e2e/e2e_test.go
```diff
@@ -0,0 +1,447 @@
...
+			Expect(podIP[:6]).Should(Equal("10.10."))
+			podIPFields := strings.Split(pod.Status.PodIP, ".")
+			field3, err := strconv.Atoi(string(podIPFields[2]))
+			Expect(err).NotTo(HaveOccurred())
+			field4, err := strconv.Atoi(string(podIPFields[3]))
+			Expect(err).NotTo(HaveOccurred())
+			if field3 <= 255 && field3 >= 0 && field4 <= 255 && field4 >= 0 {
+				expectedResult = true
+			}
+			Expect(expectedResult).Should(Equal(true))
```
maiqueb, line 317: you can always `Expect(...).To(BeTrue())` which is more idiomatic.

e2e/e2e_test.go
```diff
@@ -0,0 +1,447 @@
...
+
+			err = clientSet.CoreV1().Pods(pod.Namespace).Delete(context.TODO(), pod.Name, metav1.DeleteOptions{})
+			Expect(err).NotTo(HaveOccurred())
```
maiqueb, line 320: inline

maiqueb (changes requested):
Here's a first pass at your PR.

e2e/e2e_test.go
```diff
@@ -0,0 +1,447 @@
...
+}
+
+// DelNetAttachDef removes a net-attach-def from kubernetes
+// Returns an error variable
+func (c *ClientInfo) DelNetAttachDef(netattach *nettypes.NetworkAttachmentDefinition) error {
+	return c.NetClient.NetworkAttachmentDefinitions(netattach.ObjectMeta.Namespace).Delete(context.TODO(), netattach.Name, metav1.DeleteOptions{})
+}
+
+// Returns a network attachment definition object configured for whereabouts
+func getNetAttachDefObject() *nettypes.NetworkAttachmentDefinition {
```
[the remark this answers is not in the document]
nicklesimba: My C++ instincts make me want to disagree on the first part, but that's neither here nor there. I'll make the change. :joy:

I'm not sure about function arguments - this is a function made just for whereabouts e2e, and will always return the same net-attach-def. It'd be a perfect candidate for a const, but the compiler doesn't consider the definition I assign it to be constant, hence this function. Suggestions are welcome.

e2e/e2e_test.go
```diff
@@ -0,0 +1,447 @@
...
+	var err error
+	timeout := 30 // 30 second wait - changed to 300. will it help?
+
+	// time.Sleep(5)
+	err = WaitForPodBySelector("steady", cs, ns, WB_LABEL_EQUAL, timeout)
+	Expect(err).NotTo(HaveOccurred())
+	fmt.Println("no problems here!")
+
+	// err = waitForPodTerminateSteady(cs, ns, WB_LABEL_EQUAL, time.Duration(timeout))
+	Expect(err).NotTo(HaveOccurred())
```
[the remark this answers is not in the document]
nicklesimba: Meant to comment that out for now, good catch :sweat_smile:

e2e/e2e_test.go
```diff
@@ -0,0 +1,447 @@
...
+			num_compute_nodes int
+			kubeconfig        *string
+			LABEL             map[string]string
+			ANNOTATIONS       map[string]string
+			expectedResult    bool // maybe there is a better/cleaner way to do this.
+		)
+
+		flag.IntVar(&num_compute_nodes, "num_compute_nodes", 5, "# of compute nodes for use in test")
+
+		kubeconfig = flag.String("kubeconfig", "/tmp/kubeconfig", "(optional) absolute path to the kubeconfig file")
```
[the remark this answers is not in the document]
nicklesimba: I believe I was getting errors, something like nil pointer dereference - basically kubeconfig wasn't defined so `BuildConfigFromFlags` would result in failure. So I use this way. I'm positive you are correct, but I'm just not sure how to do your way. Suggestions are once again welcome and appreciated.

e2e/e2e_test.go
```diff
@@ -0,0 +1,447 @@
...
+
+			// dt := time.Now()
+
+			LABEL["tier"] = RS_NAME
+			ANNOTATIONS["k8s.v1.cni.cncf.io/networks"] = TEST_NETWORK_NAME
+
+			// Create a net-attach-def
+			_, err = clientInfo.AddNetAttachDef(netAttachDef)
+			Expect(err).NotTo(HaveOccurred())
+			fmt.Println("NetworkAttachmentDefinition created successfully...")
```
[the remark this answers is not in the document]
nicklesimba: By Clause is more readable if it is before an action, I will change to:
```
// Create a net-attach-def
By("creating NetworkAttachmentDefinition for whereabouts")
_, err = clientInfo.AddNetAttachDef(netAttachDef)
```

e2e/e2e_test.go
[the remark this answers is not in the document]
maiqueb: I just mean you should drop the `(optional)` from the description of the variable - afaiu, it is not optional.

e2e/e2e_test.go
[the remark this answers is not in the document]
maiqueb: I see, good point.

If you're interested in creating **always** the same net-attach-def you should have a name that indicates this is create **the** net-attach-def, not **a** net-attach-def.

The name should match what it does, something like `macvlanNetworkWithWhereaboutsIPAMNetwork`. 

I would actually add 2 functions - one that generates parametrized net-attach-defs, another that gives you **the** net-attach-def:

```golang
const (
    namespace = "ns1" // if it's other than `default` odds are you have to create / remove on the test setup / teardown
    networkName = "tenantANetwork"
)

func macvlanNetworkWithWhereaboutsIPAMNetwork()  *nettypes.NetworkAttachmentDefinition {
    networkSpec := `{
        "cniVersion": "0.3.0",
      	"disableCheck": true,
        "plugins": [
            {
                "type": "macvlan",
              	"master": "eth0",
              	"mode": "bridge",
              	"ipam": {
                    "type": "whereabouts",
                    "leader_lease_duration": 1500,
                    "leader_renew_deadline": 1000,
                    "leader_retry_period": 500,
                    "range": "10.10.0.0/16",
                    "log_level": "debug",
                    "log_file": "/tmp/wb"
              	}
            }
        ]
    }`
    return generateNetAttachDefSpec(networkName, namespace, networkSpec)
}

func generateNetAttachDefSpec(name, namespace, config string) *nettypes.NetworkAttachmentDefinition {
...
}
```

As you add more and more tests - and you require different configurations - you would add new functions that invoke the parametrized method.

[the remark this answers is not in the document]
nicklesimba: This is very forward thinking, great ideas.

.github/workflows/test.yml
```diff
@@ -65,6 +65,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       NUMBER_OF_COMPUTE_NODES: 5
+      NUMBER_OF_THRASH_ITER: 20
+      FILL_PERCENT_CAPACITY: 20
```
maiqueb, line 69: you should define these locally for the `e2e` step, below the `KUBECONFIG` environment variable - the other steps don't use this.

e2e/pod_status.go
```diff
@@ -49,6 +51,43 @@ func isPodGone(cs *kubernetes.Clientset, podName, namespace string) wait.Conditi
...
+		replicaSet, err := cs.AppsV1().ReplicaSets(namespace).Get(context.Background(), replicaSetName, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+
+		// check two things:
+		// 1. number of pods that are ready should equal that of spec
+		// 2. number of pods matching replicaSet's selector should equal that of spec
+		//    (in 0 replicas case, replicas should finish terminating before this comes true)
+		if replicaSet.Status.ReadyReplicas == (*replicaSet.Spec.Replicas) && int32(len(podList.Items)) == (*replicaSet.Spec.Replicas) {
```
maiqueb, line 70: please export this function into an helper. `isReplicaSetSynchronized(replicaSet, podList)` or something like that.

e2e/e2e_test.go
```diff
@@ -87,17 +108,202 @@ var _ = Describe("Whereabouts functionality", func() {
...
+}
+
+type envVars struct {
+	kubeconfigPath      string
+	numComputeNodes     int
+	fillPercentCapacity int
+	numThrashIter       int
+}
+
+func getEnvVars() (envVars, error) {
```
maiqueb, line 262: please avoid the `get` function name prefix - you oould name this `environment`, for instance.

e2e/e2e_test.go
```diff
@@ -87,17 +108,202 @@ var _ = Describe("Whereabouts functionality", func() {
...
+		return envVars, fmt.Errorf("must provide the number of compute nodes via the `NUMBER_OF_COMPUTE_NODES` env variable")
+	}
+	fillPercentCapacity, found := os.LookupEnv(fillCapcity)
+	if !found {
+		return envVars, fmt.Errorf("must provide the fill percent via the `FILL_PERCENT_CAPACITY` env variable")
+	}
+	numThrashIter, found := os.LookupEnv(numThrash)
+	if !found {
+		return envVars, fmt.Errorf("must provide the number of iterations via the `NUMBER_OF_THRASH_ITER` env variable")
 	}
```
maiqueb, line 287: these have defaults on the bash script - I think we can preserve that behavior. 

I would probably use defaults more suitable to running this locally - 2 nodes, 50 percent capacity, 1 iteration - but I'm also OK if you prefer to re-use the existing defaults.

e2e/e2e_test.go
```diff
@@ -169,7 +375,45 @@ func (c *ClientInfo) deletePod(pod *core.Pod) error {
...
-	if err := WaitForPodToDisappear(c.Client, pod.GetNamespace(), pod.GetName(), deleteTimeout); err != nil {
+	if err := WaitForPodToDisappear(c.Client, pod.GetNamespace(), pod.GetName(), podDeleteTimeout); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c *ClientInfo) provisionReplicaSet(replicaCount int32, label, annotations map[string]string) (*apps.ReplicaSet, error) {
+	replicaSet := replicaSetObject(replicaCount, label, annotations)
+	replicaSet, err := c.Client.AppsV1().ReplicaSets(testNamespace).Create(context.Background(), replicaSet, metav1.CreateOptions{})
```
maiqueb, line 386: nit: you could inline these:
```golang
replicaSet, err := c.Client.AppsV1().ReplicaSets(testNamespace).Create(
    context.Background(), 
    replicaSetObject(replicaCount, label, annotations), 
    metav1.CreateOptions{})
```

e2e/e2e_test.go
```diff
@@ -236,6 +480,44 @@ func podObject(label, annotations map[string]string) *core.Pod {
...
 }
 
+func replicaSetObject(replicaCount int32, label map[string]string, annotations map[string]string) *apps.ReplicaSet {
+	numReplicas := &replicaCount
+
+	return &apps.ReplicaSet{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "ReplicaSet",
+		},
```
maiqueb, line 490: nit: I think this is not needed. The kubernetes API is smart enough to know that an `apps.ReplicaSet` object has kind `ReplicaSet` . 

You can keep it if you want.

e2e/e2e_test.go
```diff
@@ -284,3 +566,39 @@ func inRange(cidr string, ip string) error {
...
+	reconcilerTimeout := defaultReconcilerTimeout
+	logLevel := "error"
+
+	logging.SetLogLevel(logLevel)
+
+	if kubeconfigPath == "" {
+		ipReconcileLoop, err = reconciler.NewReconcileLooper(context.Background(), reconcilerTimeout)
+	} else {
+		ipReconcileLoop, err = reconciler.NewReconcileLooperWithKubeconfig(context.Background(), kubeconfigPath, reconcilerTimeout)
+	}
```
maiqueb, line 583: no need for the else, right ?... you know you're not running in a pod, and as such, must pass it the path to the kubeconfig.

e2e/e2e_test.go
```diff
@@ -284,3 +566,39 @@ func inRange(cidr string, ip string) error {
...
+	if err != nil {
+		_ = logging.Errorf("failed to create the reconcile looper: %v", err)
+		os.Exit(couldNotStartOrphanedIPMonitor) // not sure if i should be exiting or returning the error?
```
maiqueb, line 586: yeah, `os.Exit` shouldn't be used. Just return an error - `return fmt.Errorf("failed to create the reconcile looper: %w", err)` .

Remove the previous logging line as well.

e2e/e2e_test.go
```diff
@@ -284,3 +566,39 @@ func inRange(cidr string, ip string) error {
...
+	}
+
+	cleanedUpIps, err := ipReconcileLoop.ReconcileIPPools(context.Background())
+	if err != nil {
+		_ = logging.Errorf("failed to clean up IP for allocations: %v", err)
+		os.Exit(failedToReconcileIPPools) // not sure if i should be exiting or returning the error?
+	}
```
maiqueb, line 593: ditto

e2e/e2e_test.go
```diff
@@ -284,3 +566,39 @@ func inRange(cidr string, ip string) error {
...
+	if len(cleanedUpIps) > 0 {
+		logging.Debugf("successfully cleanup IPs: %+v", cleanedUpIps)
+	} else {
+		logging.Debugf("no IP addresses to cleanup")
+	}
+
+	if err := ipReconcileLoop.ReconcileOverlappingIPAddresses(context.Background()); err != nil {
+		os.Exit(failedToReconcileClusterWideIPs) // not sure if i should be exiting or returning the error?
+	}
```
maiqueb, line 602: ditto

e2e/e2e_test.go
```diff
@@ -284,3 +566,39 @@ func inRange(cidr string, ip string) error {
...
+	cleanedUpIps, err := ipReconcileLoop.ReconcileIPPools(context.Background())
+	if err != nil {
+		_ = logging.Errorf("failed to clean up IP for allocations: %v", err)
+		os.Exit(failedToReconcileIPPools) // not sure if i should be exiting or returning the error?
+	}
+	if len(cleanedUpIps) > 0 {
+		logging.Debugf("successfully cleanup IPs: %+v", cleanedUpIps)
+	} else {
+		logging.Debugf("no IP addresses to cleanup")
+	}
```
maiqueb, line 598: I don't think this matters at all; I would delete these lines.

.github/workflows/test.yml
```diff
@@ -87,6 +89,6 @@ jobs:
         run: NUMBER_OF_THRASH_ITER=20 FILL_PERCENT_CAPACITY=20 ./hack/e2e-test.sh --number-of-compute $NUMBER_OF_COMPUTE_NODES
 
       - name: Execute golang based E2E tests
-        run: pushd e2e; go test -v . ; popd
+        run: pushd e2e; go test -v . -timeout 30m; popd
```
maiqueb, line 92: are you actually reading this anywhere ? From what I understood in the code below, you're just using the `defaultReconcilerTimeout`, which is 30.

e2e/e2e_test.go
```diff
@@ -87,17 +108,202 @@ var _ = Describe("Whereabouts functionality", func() {
...
+					// Check for 0 IP pool allocations
+					ipPools, err := k8sClient.ListIPPools(context.Background())
+					Expect(err).To(BeNil())
+					Expect(ipPools[0].Allocations()).To(BeEmpty()) // get ipPool by name rather than by hoping index is right - TODO
+
+					// Increase to desired number of replicas for test
+					allPods, err := clientInfo.Client.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{})
+					Expect(err).NotTo(HaveOccurred())
+					maxReplicas := (envVars.numComputeNodes*maxPodsPerNode - (len(allPods.Items))) * envVars.fillPercentCapacity / 100
+					replicaCount = int32(maxReplicas)
```
maiqueb, line 169: you could inline these 2. I don't think the `maxReplicas` is being used anywhere else - it's just an intermediate value.

e2e/e2e_test.go
```diff
@@ -87,17 +108,202 @@ var _ = Describe("Whereabouts functionality", func() {
...
+
+	By("Comparing pod ip assignments to ippool reservations")
+	// TODO - stop this ipPools[0] stuff and get it by name if there's a way to do it. low priority though
+	ipPools, err := k8sClient.ListIPPools(context.Background())
+	if err != nil {
+		return err
+	}
+
+	ipPoolAllocations := ipPools[0].Allocations() // todo, get this more elegantly than ipPools[0]
+	podList, err := ListPods(cs, testNamespace, wbLabelEqual)
```
maiqueb, line 202: You can get the correct pol by name, rather than listing all, then grabbing the first (and only ...) one.

```golang
poolName := normalizeIPRange(ipv4TestRange)
ipPool, err := k8sClient.GetIPPool(context.Background(), poolName)
```

The thing is the function `GetIPPool` does not exist yet, you'd have to create it :) 

It shouldn't be too hard; just copy https://github.com/k8snetworkplumbingwg/whereabouts/blob/3049dcec3f6a6bf90f2587e8e3a7757574fd1c49/pkg/storage/kubernetes/client.go#L79 , but using `i.client.Get(...)` instead.

You can handle that in a follow-up PR if you prefer, since this test is only using a single pool.

e2e/e2e_test.go
```diff
@@ -87,17 +108,202 @@ var _ = Describe("Whereabouts functionality", func() {
...
+
+			if reservedIP == podIP {
+				found = true
+				break
+			}
+		}
+
+		if !found {
+			return fmt.Errorf("Possible stale IP Pool: failed to find pod for IP Pool IP %s\n", allocation.IP)
+		}
```
maiqueb, line 228: you could export this into an helper func.

We could eventually think of a way to make this perform better than O(n^2), but right now, that's a premature optimization.

e2e/e2e_test.go
```diff
@@ -87,17 +108,202 @@ var _ = Describe("Whereabouts functionality", func() {
...
+				found = true
+				break
+			}
+		}
+
+		if !found {
+			fmt.Printf("possible pod IP not recorded in IP pool: failed to find IP pool allocation for pod IP %s\n", podIP)
+			return fmt.Errorf("possible pod IP not recorded in IP pool: failed to find IP pool allocation for pod IP %s\n", podIP)
+		}
+	}
```
maiqueb, line 250: likewise.

maiqueb (changes requested):
I **love** how the test reads. 

I have some minor nits, but we're close :)

e2e/e2e_test.go
```diff
@@ -87,17 +108,202 @@ var _ = Describe("Whereabouts functionality", func() {
...
+				By("deleting replicaset with whereabouts net-attach-def")
+				Expect(clientInfo.deleteReplicaSet(replicaSet)).To(Succeed())
+			})
+
+			It("allocates each IP pool entry with a unique pod IP", func() {
+				By("creating max number of pods and checking IP Pool validity")
+
+				for i := 0; i < envVars.numThrashIter; i++ {
+					dt = time.Now()
+					fmt.Printf("iteration %d of %d, %s\n", i, envVars.numThrashIter, dt.String())
```
maiqueb, line 147: remove the `fmt.Printf` calls.

e2e/e2e_test.go
nicklesimba: I like the idea of default config being suited for local runs.

e2e/e2e_test.go
nicklesimba: I'll remove it, but I won't resolve this conversation until I check that nothing breaks, just to be safe.

.github/workflows/test.yml
maiqueb: my bad. 

I agree, 30 minutes can be too optimistic. I would raise it to 60.

maiqueb (changes requested):
Furthermore, once this test is OK, you should remove the original bash script.

e2e/e2e_test.go
nicklesimba: I would be interested in hearing how this could perform better than O(n^2) :thinking:

.github/workflows/test.yml
nicklesimba: Lol I have since upped it to 2 hours. But eventually we should make it as short as possible, of course.

e2e/e2e_test.go
```diff
@@ -32,7 +32,7 @@ import (
 const (
 	wbLabelEqual     = "tier=whereabouts-scale-test"
 	testNamespace    = "kube-system"
-	rsCreateTimeout  = 30 * time.Second
+	rsCreateTimeout  = 600 * time.Second
```
maiqueb, line 35: well, I don't understand how this fixed the timeout - apparently the traces were just very misleading. 

Never the less, I think 600 seconds is waaaay too much. Could you try it with 1 or 2 minutes please ? Seems more sensible.

maiqueb:
@nicklesimba I think you can remove the `DoNotMerge` tag from the PR title. 

I'll give it one more round of reviews, but, from my perspective, all this PR is missing is removing everything related to the bash script (the script itself + the github action).

Just to have some figures:
  - e2e bash script completed in:   43m 33s 
  - e2e golang tests completed in: 38m 48s
  
  The golang based tests furthermore feature one additional test (takes ~= 15 seconds ...).

Very good work.

e2e/e2e_test.go
```diff
@@ -42,12 +50,19 @@ var _ = Describe("Whereabouts functionality", func() {
 	Context("Test setup", func() {
 		var (
 			clientInfo   *ClientInfo
+			config       *rest.Config
+			envVars      envVars
+			err          error
```
maiqueb, line 55: I would also scope err locally whenever it is required.

e2e/e2e_test.go
```diff
@@ -81,23 +97,253 @@ var _ = Describe("Whereabouts functionality", func() {
...
 				Expect(inRange(ipv4TestRange, secondaryIfaceIP)).To(Succeed())
 			})
 		})
+
+		Context("Replicaset tests", func() {
+
+			var (
+				replicaCount int32
+				k8sClient    *kubeClient.Client
+				label        string
```
maiqueb, line 114: I don't understand this variable. Why don't you use directly `wbLabelEqual` instead whenever that is required ?

e2e/e2e_test.go
```diff
@@ -81,23 +97,253 @@ var _ = Describe("Whereabouts functionality", func() {
...
+				ipPool, err := k8sIPAM.GetIPPool(context.Background(), ipPoolName)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(ipPool.Allocations()).To(BeEmpty())
+			})
+		})
 	})
 })
 
-func clusterConfig() (*rest.Config, error) {
+func iPPoolConsistency(kubeconfigPath string, k8sClient *kubeClient.Client, k8sIPAM *kubeClient.KubernetesIPAM, cs *kubernetes.Clientset) error {
```
maiqueb, line 208: afaiu, the `k8sClient` attribute is not being used.

e2e/e2e_test.go
```diff
@@ -81,23 +97,253 @@ var _ = Describe("Whereabouts functionality", func() {
...
+				replicaCount int32
+				k8sClient    *kubeClient.Client
+				label        string
+				k8sIPAM      *kubeClient.KubernetesIPAM // - may be used for GetIPPools - todo
+			)
+
+			BeforeEach(func() {
+				By("creating a replicaset with whereabouts net-attach-def")
+
+				replicaCount = int32(0) // begin test with 0 replicas.
```
maiqueb, line 121: I would go as far as to expose a constant in the `"Replicaset tests"` context:
- emptyReplicaSet / noReplicas = 0

and use that instead. No need for the `replicaCount` variable at all.

e2e/e2e_test.go
nicklesimba: Let's look at the formula for how many replicas should be created.

```replicaCount = int32((envVars.numComputeNodes*maxPodsPerNode - (len(allPods.Items))) * envVars.fillPercentCapacity / 100)```

Let's say we have 80% fill capacity set for our e2e test. We have 5 compute nodes currently, and we know there's 110 pods per node maximum. That's about 550 pods, and 80% of that is 440 pods. Let's say the cluster has 40 pods prior to the e2e test, maybe doing whereabouts or multus backend work. This still totals 400 pods. 

I'm not sure 1 or 2 minutes is enough for this many replicas to come up. On my local environment, it takes several minutes to bring up just 100 pods. Let me know if I've made a mistake, happy to be proven wrong, but I think 10 minutes is reasonable for now.

File: e2e/e2e_test.go [hunk 30]
```diff
@@ -81,23 +97,253 @@ var _ = Describe("Whereabouts functionality", func() {
...
+
+					// Check for 0 IP pool allocations
+					ipPool, err := k8sIPAM.GetIPPool(context.Background(), ipPoolName)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(ipPool.Allocations()).To(BeEmpty())
+
+					// Increase to desired number of replicas for test
+					allPods, err := clientInfo.Client.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{})
+					Expect(err).NotTo(HaveOccurred())
+					replicaCount = int32((envVars.numComputeNodes*maxPodsPerNode - (len(allPods.Items))) * envVars.fillPercentCapacity / 100)
```
maiqueb, line 174: instead of using this, I would create a function that returns this, and inline the usage. The trick here is the proper naming, for instance:

```golang
replicaSet, err = clientInfo.updateReplicaSet(replicaSetObject(
                                 envVars.maxReplicas(allPods.Items),
                                 podTierLabel(rsName),
                                 podNetworkSelectionElements(testNetworkName),
))

func (v envVars) maxReplicas(allPods []v1.Pod) int32 {
    const maxPodsPerNode  = 110
    return int32(
        (v.numComputeNodes*maxPodsPerNode - (len(allPods))) * v.fillPercentCapacity / 100)
}
```

e2e/e2e_test.go
```diff
@@ -81,23 +97,253 @@ var _ = Describe("Whereabouts functionality", func() {
...
+			})
+
+			AfterEach(func() {
+				By("deleting replicaset with whereabouts net-attach-def")
+				Expect(clientInfo.deleteReplicaSet(replicaSet)).To(Succeed())
+			})
+
+			It("allocates each IP pool entry with a unique pod IP", func() {
+				const (
+					maxPodsPerNode  = 110
```
maiqueb, line 149: this could be local to the func that calculates the maximum number of replicas.

e2e/e2e_test.go
```diff
@@ -81,23 +97,253 @@ var _ = Describe("Whereabouts functionality", func() {
...
+					podNetworkSelectionElements(testNetworkName),
+				))
+				Expect(err).NotTo(HaveOccurred())
+
+				Expect(WaitForReplicaSetSteadyState(clientInfo.Client, testNamespace, label, replicaSet, rsSteadyTimeout)).To(Succeed())
+
+				// Check for 0 IP pool allocations
+				ipPool, err := k8sIPAM.GetIPPool(context.Background(), ipPoolName)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(ipPool.Allocations()).To(BeEmpty())
```
maiqueb, line 202: this clearly belongs on the test teardown (after each section).

I also don't understand why you scale down the replica set. Shouldn't you just remove it ?... Without it, the replicaset leaks - but has no pods.

Furthermore, check if you can encapsulate this into a neat function.

e2e/e2e_test.go
```diff
@@ -81,23 +97,253 @@ var _ = Describe("Whereabouts functionality", func() {
...
+		return envVars, err
+	}
+	envVars.fillPercentCapacity, err = strconv.Atoi(fillPercentCapacity)
+	if err != nil {
+		return envVars, err
+	}
+	envVars.numThrashIter, err = strconv.Atoi(numThrashIter)
+	if err != nil {
+		return envVars, err
 	}
```
maiqueb, line 341: I would actually create an helper each for compute nodes, percent capacity, and num iterations, then use it.

For instance:
```golang
func environment() (envVars, error) {
    ...
    numComputeNodes, err := computeNodes()
    if err != nil {
        return envVars{}, err
    }
    ...
    return envVars{
        numComputeNodes: numComputeNodes,
        ...
    }
}

func computeNodes() (int, error) {
        numComputeNodes, found := os.LookupEnv(numCompute)
	if !found {
		numComputeNodes = "2"
	}
        return strconv.Atoi(numComputeNodes)
}
```

Furthermore, `return envVars, err` is not as robust as it could be - imagine the caller forgets to check the error. He would have an half created env vars, which surely would be hard to debug. 

You can make the code more resilient for returning `nil, err` - you'd have to adjust the return type, making it a pointer. Or you could alternatively return an empty envVars struct.

e2e/e2e_test.go
```diff
@@ -81,23 +97,253 @@ var _ = Describe("Whereabouts functionality", func() {
...
+				k8sIPAM      *kubeClient.KubernetesIPAM // - may be used for GetIPPools - todo
+			)
+
+			BeforeEach(func() {
+				By("creating a replicaset with whereabouts net-attach-def")
+
+				replicaCount = int32(0) // begin test with 0 replicas.
+				label = wbLabelEqual
+
+				k8sClient, err = kubeClient.NewClientViaKubeconfig(envVars.kubeconfigPath)
```
maiqueb, line 124: This is failing when compiling since pr #209 was merged. You now have to pass it also a timeout.

I would recommend 10 seconds - `10*time.Second` - passed via a const.

e2e/e2e_test.go
maiqueb: Pff. You might be right. Let's be safe then.

e2e/e2e_test.go
nicklesimba: Moved to `AfterEach`. :+1:  

I can't just remove the replicaset and do nothing else however. After the test finishes, the replicas may still be in `Terminating` state. This will effect Martin's test now, and maybe in the future someone else's tests. While this does add a little bit of code, it guarantees 0 IP pool allocations after the test runs.

I put all instances of this kind of code in a neat function like you suggested, as well.

maiqueb: Yes, that is correct: you should delete the replica set **and** wait for the pods to be gone.

e2e/e2e_test.go
```diff
@@ -446,22 +445,22 @@ func (c *ClientInfo) deleteReplicaSet(replicaSet *apps.ReplicaSet) error {
...
 		emptyReplicaSet = 0
 		rsSteadyTimeout = 2 * rsCreateTimeout
 	)
 	var err error
 
 	// Set to 0 replicas
-	replicaSet, err = clientInfo.updateReplicaSet(replicaSetObject(
+	replicaSet, err := clientInfo.updateReplicaSet(replicaSetObject(
 		emptyReplicaSet,
 		podTierLabel(rsName),
```
maiqueb, line 465: this is basically using a constant defined in https://github.com/k8snetworkplumbingwg/whereabouts/pull/181/commits/f9481dded3a92dec83d26fbefafdd0e3dc2f3506#diff-e91d29c6d4c58503d9c2d72e2b04f1b70901a613e23008d1d1cfe8beff7038d4R40

Please pass the name (and namespace) of the replicaset as an input parameter. We want this to work for more than one scenario.

e2e/e2e_test.go
```diff
@@ -169,9 +398,90 @@ func (c *ClientInfo) deletePod(pod *core.Pod) error {
...
+
+// Waits for all replicas to be fully removed from replicaset, and checks that there are 0 ip pool allocations
+func checkZeroIPPoolAllocationsAndReplicas(clientInfo *ClientInfo, k8sIPAM *kubeClient.KubernetesIPAM) error {
+	const (
+		emptyReplicaSet = 0
+		rsSteadyTimeout = 2 * rsCreateTimeout
+	)
+	var err error
+
+	// Set to 0 replicas
```
maiqueb, line 462: nit: ditch this comment; by using below the `emptyReplicaSet` const, it's pretty obvious what you're doing.

e2e/e2e_test.go
```diff
@@ -169,9 +398,90 @@ func (c *ClientInfo) deletePod(pod *core.Pod) error {
...
+	}
+
+	// Check for 0 IP pool allocations
+	ipPool, err := k8sIPAM.GetIPPool(context.Background(), ipPoolName)
+	if err != nil {
+		return err
+	}
+
+	if len(ipPool.Allocations()) != 0 {
+		return fmt.Errorf("Expected 0 IP pool allocations, got %d instead", len(ipPool.Allocations()))
```
maiqueb, line 483: I'm seeing this fail in the [e2e tests](https://github.com/k8snetworkplumbingwg/whereabouts/runs/5693681338?check_suite_focus=true#step:7:47); I think it would help somehow to know in which iteration did the test fail.

How about letting the called indicate that ? Something like (in line [152](https://github.com/k8snetworkplumbingwg/whereabouts/pull/181/files#diff-e91d29c6d4c58503d9c2d72e2b04f1b70901a613e23008d1d1cfe8beff7038d4R152)):
```golang
Expect(checkZeroIPPoolAllocationsAndReplicas(clientInfo, k8sIPAM)).To(Succeed(), fmt.Sprintf("Stale IP allocations in iteration: %d", i))
```

nicklesimba: Miguel, for the record, I somehow missed a lot of these comments from 10 days ago. Apologies! I'm catching up on these, but for this comment I think (for now) the problem is solved. I created a WaitForZeroIPPoolAllocations() function that gives kubernetes adequate time to clean up. I'm going to leave this comment unresolved, because it's a great idea in the event the problem resurfaces.

e2e/e2e_test.go
```diff
@@ -81,23 +101,232 @@ var _ = Describe("Whereabouts functionality", func() {
...
+		return err
+	}
+
+	if err := foundPodIPforAllIPPoolIPs(ipPoolAllocations, podList); err != nil {
+		return err
+	}
+
+	if err := foundIPPoolIPforAllPodIPs(ipPoolAllocations, podList); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
nicklesimba: Sure.

e2e/e2e_test.go
```diff
@@ -48,12 +57,22 @@ var _ = Describe("Whereabouts functionality", func() {
 	Context("Test setup", func() {
 		var (
 			clientInfo   *ClientInfo
+			envVars      *envVars
```
maiqueb, line 60: the variable name is shadowing the variable type - please rename the variable type to something else - `testConfig` for instance.

e2e/e2e_test.go
```diff
@@ -26,17 +27,25 @@ import (
...
 const (
-	testNetworkName = "wa-nad"
-	testNamespace   = "default"
-	testImage       = "quay.io/dougbtv/alpine:latest"
-	ipv4TestRange   = "10.10.0.0/16"
-	singlePodName   = "whereabouts-basic-test"
-	createTimeout   = 10 * time.Second
-	deleteTimeout   = 2 * createTimeout
+	wbLabelEqual     = "tier=whereabouts-scale-test"
+	testNamespace    = "kube-system"
```
maiqueb, line 38: I think we should use a different namespace for the tests: this is where the kubernetes infrastructure pods live.

I am OK with using the `default` namespace.

A follow up PR should generate a new namespace for each test, to ensure no collisions on the tests.

maiqueb, line 38: This is why the stateful set pods stopped being ready: their pods are looking for the network-attachment-definition in the default namespace. 

I recommend you use that namespace as well.

maiqueb (approved):
I think this is good enough. 

More improvements can happen in follow-up PRs.

Thanks for the patience, and good work.

maiqueb:
Would you mind squashing the 3 commits into one ?

nicklesimba:
The commit history is extremely messy under the hood, which is more than likely my own fault. I'm going to close this PR shortly and open a new one with the changes added as a single commit - since they're approved here anyway. I'll reference and link this PR in the new one. This way, context is provided to anyone who needs it.

maiqueb:
Abandoned in lieu of #215 .

PR #183, opened 2022-02-03
build: synch the golang builders w/ the go version on go.mod

maiqueb:
go.mod has 1.15 while the builders have 1.13.

Oddly enough, for whatever reason, the openshift img builder already runs 1.15.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.13
+FROM golang:1.15
 ADD . /usr/src/whereabouts
 RUN mkdir -p $GOPATH/src/github.com/k8snetworkplumbingwg/whereabouts
 WORKDIR $GOPATH/src/github.com/k8snetworkplumbingwg/whereabouts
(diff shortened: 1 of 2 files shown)


s1061123:
If we change the version, we should update up to 1.16 because golang community supports 1.16+

https://endoflife.date/go

or
https://go.dev/doc/devel/release (please see 'Release Policy' section)

maiqueb:
> If we change the version, we should update up to 1.16 because golang community supports 1.16+
> 
> https://endoflife.date/go
> 
> or https://go.dev/doc/devel/release (please see 'Release Policy' section)

The goal of this PR is to align whatever is specified on the go module file with whatever is used to build the code; it is not to bump the golang version to the supported version. 

I think your suggestion is required - I just think it should be done in follow up PR(s).

dougbtv (approved):

maiqueb:
Oh since #184 was already merged, no point merging this one.

I'll just abandon it.

Thanks.

PR #184, opened 2022-02-04
Bump golang 1.16

maiqueb:
Depends-on: #183

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,7 +5,7 @@ jobs:
     strategy:
       matrix:
         #go-version: [1.15.x, 1.16.x]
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -5,7 +5,7 @@ jobs:
     strategy:
       matrix:
         #go-version: [1.15.x, 1.16.x]
-        go-version: [1.15.x]
```
maiqueb, line 8: I wonder why this is not testing the 2 versions .... Is there a reason for it ? I.e. what's the point of building for the 2 versions if we just test one of them ?

dougbtv (approved):

PR #186, opened 2022-02-04
Fix context reconciler

maiqueb:
Follow the golang context convention [0], which says:
"""
Do not store Contexts inside a struct type; instead, pass a Context explicitly
to each function that needs it. The Context should be the first parameter,
typically named ctx:
...
"""

[0] - https://pkg.go.dev/context

Furthermore, this PR also changes the reconciler behavior to use a different
context per each list operation. All contexts are cancelable, and have a timeout.
    
The timeout can be specified via the reconciler command line arguments,
and defaults to 30 seconds.

Fixes: #172

### DIFF
--- a/cmd/reconciler/ip.go
+++ b/cmd/reconciler/ip.go
@@ -25,14 +25,14 @@ func main() {
 	if kubeConfigFile == nil {
 		ipReconcileLoop, err = reconciler.NewReconcileLooper(ctx)
 	} else {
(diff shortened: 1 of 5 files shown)


maiqueb:
@xagent003 / @dougbtv would you please review this PR ?

xagent003:
Looks good to me

/approve
/lgtm

dougbtv (approved):
LGTM!

PR #193, opened 2022-02-11
WIP: ip-reconciler should have known errors from api server on instantiation

dougbtv:


cmd/reconciler/ip.go
```diff
@@ -5,11 +5,30 @@ import (
 	"flag"
 	"github.com/k8snetworkplumbingwg/whereabouts/pkg/storage"
 	"os"
+	"strings"
 
 	"github.com/k8snetworkplumbingwg/whereabouts/pkg/logging"
 	"github.com/k8snetworkplumbingwg/whereabouts/pkg/reconciler"
 )
 
+func knownErrorCase(err error) bool {
```
s1061123, line 14: we may have comment for the function to mention that it is used for accept some failure case as valid case.

dougbtv: Agreed, added a comment :+1:

dougbtv:
> all these logs will be gone on successes.
> 
> IMO, if we take this route, we need to preserve one successful pod around.

Nice catch. Yes -- I think this is critical and actually buys a lot for the tradeoff we have with losing visibility for a "known error" that we should be catching.

Updated the cronjob yaml

maiqueb (approved):

cmd/reconciler/ip.go
```diff
@@ -29,13 +49,21 @@ func main() {
 	}
 	if err != nil {
 		_ = logging.Errorf("failed to create the reconcile looper: %v", err)
-		os.Exit(couldNotStartOrphanedIPMonitor)
+		if knownErrorCase(err) {
+			os.Exit(0)
+		} else {
+			os.Exit(couldNotStartOrphanedIPMonitor)
+		}
```
maiqueb, line 56: would it make sense to add another helper func here ? This code is clearly repeated 3 times. I propose something like:

```golang
func tolerateKnownErrorCasesOrFail(err error, errorCodeToReturn int) {
    const pretendEverythingIsFine = 0

    if knownErrorCase(err) {
        os.Exit(pretendEverythingIsFine)
    } else {
        os.Exit(errorCodeToReturn)
    }
}
```

maiqueb (commented):
This is an oppinionated nit you can choose to ignore.

cmd/reconciler/ip.go
dougbtv: Implemented! Good eye, agreed.

cmd/reconciler/ip.go
```diff
@@ -28,21 +63,22 @@ func main() {
 	}
 	if err != nil {
 		_ = logging.Errorf("failed to create the reconcile looper: %v", err)
```
nicklesimba, line 65: The `_ = logging.Error()` syntax might be throwing me off... but I think this still causes an output logging print -- and just throws away the return value.

In the case that we tolerate a known error, do we want to log (and therefore print) the error twice (once to state the error, and a second time to say it's a known error)? Alternatively, we could remove this line and move it to the `else` condition of `tolerateKnownErrorCasesOrFail`, which would mean we always print the error once.

cmd/reconciler/ip.go
```diff
@@ -28,21 +63,22 @@ func main() {
...
-		os.Exit(couldNotStartOrphanedIPMonitor)
+		tolerateKnownErrorCasesOrFail(err, couldNotStartOrphanedIPMonitor)
 	}
 
 	cleanedUpIps, err := ipReconcileLoop.ReconcileIPPools(context.Background())
 	if err != nil {
 		_ = logging.Errorf("failed to clean up IP for allocations: %v", err)
```
nicklesimba, line 71: [Ditto](https://www.youtube.com/watch?v=ODKTITUPusM)

cmd/reconciler/ip.go
```diff
@@ -28,21 +63,22 @@ func main() {
...
+
 	if len(cleanedUpIps) > 0 {
 		logging.Debugf("successfully cleanup IPs: %+v", cleanedUpIps)
 	} else {
 		logging.Debugf("no IP addresses to cleanup")
 	}
 
 	if err := ipReconcileLoop.ReconcileOverlappingIPAddresses(context.Background()); err != nil {
-		os.Exit(failedToReconcileClusterWideIPs)
+		tolerateKnownErrorCasesOrFail(err, failedToReconcileClusterWideIPs)
```
nicklesimba, line 82: To add to my previous point, if the error here is not a known error case, the error will not ever be logged. If we know that the error here is guaranteed to be a known case, this is perfectly fine.

cmd/reconciler/ip.go
```diff
@@ -4,13 +4,48 @@ import (
...
+	"strings"
 
 	"github.com/k8snetworkplumbingwg/whereabouts/pkg/logging"
 	"github.com/k8snetworkplumbingwg/whereabouts/pkg/reconciler"
 )
 
 const defaultReconcilerTimeout = 30
 
+// Matches known error cases from which the process should exit cleanly.
+func knownErrorCase(err error) bool {
```
danielmellado, line 16: It may be interesting to actually store this out around in some var. Like, setting a threshold of where this actually may be hitting a real issue sounds reasonable to me. Like a timeout may be on the infra, multiple ones may not be.

dougbtv: Hey Daniel, that's a nice suggestion. What I've done is updated this to cycle through a constant set of known errors for now, and potentially in a future edition make this configurable. Thanks!

cmd/reconciler/ip.go
dougbtv: This is a really nice catch, thanks! Definitely need an error logged here.

cmd/reconciler/ip.go
dougbtv: Solidly done. I went ahead and made it so that the error message is passed in as a parameter to the `tolerateKnownErrorCasesOrFail()` method, and now it should only be logged once-per-each instead of a double logging in the case of a known error

maiqueb (commented):
@dougbtv I can't say I like sweeping this under the rug, but I honestly don't know what else to do on such situations.

cmd/reconciler/ip.go
```diff
@@ -4,13 +4,52 @@ import (
...
+		"connection refused",
+	}
+}
+
+// Matches known error cases from which the process should exit cleanly.
+func knownErrorCase(err error) bool {
+
+	knownerrs := knownErrors()
+
+	for _, knownerr := range knownerrs {
```
s1061123, line 32: ditto, but it could be without '()' if you use variable, not function.

```
for _, knownerr := range knownErrors
```

s1061123 (changes requested):

dougbtv:
too hacky.

Issue #182, opened 2022-02-02
ip-reconciler issue with StatefullSet cleanup


xagent003:
@maiqueb @dougbtv There is a larger problem with StatefulSets and container IDs. My current PR (use the container ID to match when deleting) only fixes one issue https://github.com/k8snetworkplumbingwg/whereabouts/pull/180

A short while back we changed the isPodAlive() to skip pods in Pending phase: https://github.com/k8snetworkplumbingwg/whereabouts/commit/55be906b778f95d2e907fe06d2b1b464f78cf2b8

This was because if the ip-reconciler ran while a pod was being deployed, it cleaned up the reservation because kubelet takes some time before adding the network-status annotations and IPs to the Pod resource. So the Pod resource is missing the IP temporarily, and the reservation would be erroneously marked for cleanup.

Now there is new issue where, if the IP range is full. Suppose 3 IPs in range, 3 Pods exist. Now suppose an STS Pod gets recreated after a node goes down. It hits ContainerCreating state and stays there because No IP in Range.

When ip-reconciler runs, it only checks by podRef. and it will match because it's a StatefulSet. The Pod that is stuck ContainerCreating will also be in Pending Phase, as kubelet is still retrying/awaiting the CNI. Therefore the orhpaned IP won't get cleaned up, and new Pod can never come up.

```
func (rl ReconcileLooper) isPodAlive(podRef string, ip string) bool {
	for livePodRef, livePod := range rl.liveWhereaboutsPods {
		if podRef == livePodRef {
			livePodIPs := livePod.ips
			logging.Debugf(
				"pod reference %s matches allocation; Allocation IP: %s; PodIPs: %s",
				livePodRef,
				ip,
				livePodIPs)
			_, isFound := livePodIPs[ip]
			return isFound || livePod.phase == v1.PodPending
		}
	}
	return false
```

How can we match by the containerID in the IPPool? Except the containerID stored in the reservation is the pause container ID. And this container ID is nowhere to be found in the k8s Pod resource. And I couldn't find an easy way to get the mapping (without going on actual host, grepping output of docker ps, etc...). So not sure how the ip-reconciler could achieve this. Meaning we are stuck using the podRef... which isn't unique for StatefulSets.

If we remove the PodPending check, we are back to issue #162: https://github.com/k8snetworkplumbingwg/whereabouts/issues/162

But this seems less likely to occur than the issue described in this ticket.

Issue #162, opened 2021-11-08
ip-reconciler race condition with Whereabouts, leads to IP cleanup and duplicate IPs


xagent003:
I'm seeing another duplicate IP issue caused by an accidental IP cleanup, it appears the ip-reconciler CronJob has a race condition with whereabouts allocating IPs. I see whereabouts allocates a certain IP to a Pod, Patches the IPPool CR, and updates resourceversion... all good so far.

However, if the ip-reconciler Job runs at the right time (Change the schedule to 1 minute, to increase the odds), it gets the updated IPPool with new allocation. But when it fetches the Pod, the IP has not yet been updated on the Pod. This is maybe due to kubelet or k8s still processing the container and not adding the network status w/ IPs and macs. So the ip-reconciler sends a "remove" Patch operation.

We can see that whereabouts on a node updated the Pool with allocation. I added my own logs in storage/kubernetes/ipam.go to print the CR to get the ResourceVersion and Spec. We can see it allocated for Pod web-2:

```

2021-11-06T03:38:11Z [debug] PF9: re-getting IPPool...
[...]

PR #195, opened 2022-02-16
Rechecking pending Pods

xagent003:
This fixes https://github.com/k8snetworkplumbingwg/whereabouts/issues/182

while hopefully not re-breaking https://github.com/k8snetworkplumbingwg/whereabouts/issues/162

We shouldn't treat all pending Pods as "alive" and skip the check. The list of all Pods fetch'd earlier may be stale, and as observed in some scenarios, several seconds before the ip-reconciler does the isPodAlive check.

Instead, can we retry a Get on an individual Pod, with the hopes that it has final IP/network annoations? So refetch the Pod, up to 5 times, if it is Pending state. After that, just do the IP matching check like before


xagent003:
@dougbtv and @maiqueb Can you take a look. I did verify that #162 did not re-break:

```
xagent@pop-os:~/luigi/samples/macvlan$ kubectl logs ip-reconciler-27417868-ldhrq
2022-02-17T04:28:03Z [debug] NewReconcileLooper - Kubernetes config file located at: /host/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig
2022-02-17T04:28:04Z [debug] successfully read the kubernetes configuration file located at: /host/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig
2022-02-17T04:28:04Z [debug] listing IP pools
2022-02-17T04:28:04Z [debug] Added IP 10.128.165.32 for pod default/web-0
2022-02-17T04:28:04Z [error] could not parse network annotation  for pod: default/web-1; error: unexpected end of JSON input
2022-02-17T04:28:04Z [debug] the IP reservation: IP: 10.128.165.32 is reserved for pod: default/web-0
2022-02-17T04:28:04Z [debug] pod reference default/web-0 matches allocation; Allocation IP: 10.128.165.32; PodIPs: map[10.128.165.32:{}]
2022-02-17T04:28:04Z [debug] the IP reservation: IP: 10.128.165.33 is reserved for pod: default/web-1
2022-02-17T04:28:04Z [debug] pod reference default/web-1 matches allocation; Allocation IP: 10.128.165.33; PodIPs: map[]
2022-02-17T04:28:04Z [debug] Re-fetching Pending Pod: default/web-1 IP-to-match: 10.128.165.33
2022-02-17T04:28:05Z [debug] Added IP 10.128.165.33 for pod default/web-1
2022-02-17T04:28:05Z [debug] Got refreshed pod: &{map[10.128.165.33:{}] Pending}
2022-02-17T04:28:05Z [debug] pod reference default/web-1 matches allocation; Allocation IP: 10.128.165.33; PodIPs: map[10.128.165.33:{}]
2022-02-17T04:28:05Z [debug] Pod now has IP annotation while in Pending
2022-02-17T04:28:05Z [debug] pod reference default/web-0 matches allocation; Allocation IP: 10.128.165.32; PodIPs: map[10.128.165.32:{}]
2022-02-17T04:28:05Z [debug] pod reference default/web-1 matches allocation; Allocation IP: 10.128.165.33; PodIPs: map[]
2022-02-17T04:28:05Z [debug] Re-fetching Pending Pod: default/web-1 IP-to-match: 10.128.165.33
2022-02-17T04:28:05Z [debug] Added IP 10.128.165.33 for pod default/web-1
2022-02-17T04:28:05Z [debug] Got refreshed pod: &{map[10.128.165.33:{}] Pending}
2022-02-17T04:28:05Z [debug] pod reference default/web-1 matches allocation; Allocation IP: 10.128.165.33; PodIPs: map[10.128.165.33:{}]
2022-02-17T04:28:05Z [debug] Pod now has IP annotation while in Pending
2022-02-17T04:28:05Z [debug] no IP addresses to cleanup
xagent@pop-os:~/luigi/samples/macvlan$ kubectl get ippools.whereabouts.cni.cncf.io 10.128.165.0-24 -o yaml
apiVersion: whereabouts.cni.cncf.io/v1alpha1
kind: IPPool
metadata:
  creationTimestamp: "2022-02-16T07:41:55Z"
  generation: 90
  managedFields:
  - apiVersion: whereabouts.cni.cncf.io/v1alpha1
    fieldsType: FieldsV1
    fieldsV1:
      f:spec:
        .: {}
        f:allocations:
          .: {}
          f:32:
            .: {}
            f:id: {}
            f:podref: {}
          f:33:
            .: {}
            f:id: {}
            f:podref: {}
          f:34:
            .: {}
            f:id: {}
            f:podref: {}
        f:range: {}
    manager: whereabouts
    operation: Update
    time: "2022-02-17T04:28:08Z"
  name: 10.128.165.0-24
  namespace: default
  resourceVersion: "183433"
  uid: 41f5ab5f-2b34-4a5a-94ba-71c604c0a1d2
spec:
  allocations:
    "32":
      id: f7da88514327587cdd23fb0c5d7596341d2d422dea75fd7650c82fd229eb8051
      podref: default/web-0
    "33":
      id: 286db456fecf6f516721347ba004e82a823e9e5768984340ff3b44fdfa0f3c67
      podref: default/web-1
    "34":
      id: 919134bbeb8a0b1128724bf280184b48d3f3bd35aaa7bae8727324b259cb11fb
      podref: default/web-2
  range: 10.128.165.0/24
```

As you can see, Pod web-1 was missing IP annotation and in Pending phase. I refreshed the Pod, this time it had the IP annotation the first time around, and found a match. So no bad cleanup happened.

xagent003:
Ok, I verified that it also fixes #182 

Shut down a node with a STS and network with full IPs. Force deleted the Pod, so it gets rescheduled. It remains in ContainerCreating state with "Could not allocate IP in range: ip: 10.128.165.32 / - 10.128.165.33" error

Then the reconciler runs. It cleans up the stale IP (belonging to the Pod I force deleted from the dead node), despite the new replica with same podRef being in ContainerCreating state aka Pending phase. Now the new replica can create and moves to Running

seen in ip-reconciler logs:

```
2022-02-17T07:01:05Z [debug] Going to update the reserve list to: [IP: 10.128.165.32 is reserved for pod: default/web-0]
2022-02-17T07:01:05Z [debug] PF9: patch = [{remove /spec/allocations/33 <nil>}]
2022-02-17T07:01:05Z [debug] PF9: wrote patchdata: [{"op":"test","path":"/metadata/resourceVersion","value":"204380"},{"op":"remove","path":"/spec/allocations/33"}]
2022-02-17T07:01:05Z [debug] successfully cleanup IPs: [10.128.165.33]
```

verified the IP Pool shows old reservation cleaned up vs new one:

```
spec:
  allocations:
    "32":
      id: 1e62cb66ff77fbf409633ca4cef7b2b641427c347eb2269cd23efaaba0bb1e00
      podref: default/web-0
    "33":
      id: 40d15102db0b75cad2faff2ccee483bd4e6c25fdd477af9caa2e97b71423b5f7
      podref: default/web-1
  range: 10.128.165.0/24
spec:
  allocations:
    "32":
      id: 1e62cb66ff77fbf409633ca4cef7b2b641427c347eb2269cd23efaaba0bb1e00
      podref: default/web-0
    "33":
      id: 3d6225ea764ed6e3d5650110725c62aa41a22c935c9949ef1572b03c1f7e6bfd
      podref: default/web-1
  range: 10.128.165.0/24
```

pkg/reconciler/iploop.go
```diff
@@ -101,19 +102,76 @@ func (rl *ReconcileLooper) findOrphanedIPsPerPool(ipPools []storage.IPPool) erro
...
+					} else {
+						isFound = checkIpOnPod(podToMatch, podRef, ip)
+						// Short-circuit - Pending Pod may have IP now
+						if isFound {
+							logging.Debugf("Pod now has IP annotation while in Pending")
+							return true
+						}
+					}
+				}
+				isFound = checkIpOnPod(podToMatch, podRef, ip)
```
maiqueb, line 134: Would it make sense to use [exponential backoff](https://github.com/kubernetes/apimachinery/blob/ddfe1eccd72c0752f52f4c55845ea9fa73809cca/pkg/util/wait/wait.go#L419) here ? 

Also, I think this could be re-written with k8s primitives:


```golang
import "k8s.io/apimachinery/pkg/util/wait"

timeout      := time.Second
pollInterval := 200*time.Millisecond  // 5 * 0.2 = 1

// PollImmediate tries a condition func until it returns true, an error, or the timeout
// is reached.
// ... 
waitErr := wait.PollImmediate(pollInterval, timeout, func() (bool, error) {
    pod := rl.refreshPod(livePodRef)

    if podToMatch == nil {
        logging.Debugf("Cleaning up...")
        return false, fmt.Errorf("pod is gone")
    } else if isPodPending(pod) {
        logging.Debugf("Pending Pod is now in phase: %s", podToMatch.phase)
        return true, nil
    } else {
        return isIPonPod(pod, podRef, ip), nil // I'm not entirely sure I did not oversimplify this ... 
    }
})
```

pkg/reconciler/iploop.go
```diff
@@ -101,19 +102,76 @@ func (rl *ReconcileLooper) findOrphanedIPsPerPool(ipPools []storage.IPPool) erro
...
+	}
+
+	wrappedPod := wrapPod(*pod)
+	logging.Debugf("Got refreshed pod: %v", wrappedPod)
+	return wrappedPod
+}
+
+func splitPodRef(podRef string) (string, string) {
+	var namespace string
+	var podName string
```
maiqueb, line 163: nit: I don't see a reason for declaring the variables here, you can just `:=` below.

Or even - once you know `len(namespacedName) == 2` - return directly: `return namespacedName[0], namespacedName[1]`.

not sure what I wrote is easy to understand, I mean something like:
```golang
func splitPodRef(podRef string) (string, string) {
    namespacedName := strings.Split(podRef, "/")
    if len(namespacedName) != 2 {
        logging.Errorf("Failed to split podRef %s", podRef)
        return "", ""
    }

    return namespacedName[0], namespacedName[1]
}
```

pkg/storage/storage.go
```diff
@@ -13,7 +13,8 @@ var (
 	RequestTimeout = 10 * time.Second
 
 	// DatastoreRetries defines how many retries are attempted when updating the Pool
-	DatastoreRetries = 100
+	DatastoreRetries  = 100
+	PodRefreshRetries = 5
```
maiqueb, line 17: this might not be needed if we end up using `wait.PollImmediate` or `wait.ExponentialBackoff` .

pkg/storage/kubernetes/client.go
```diff
@@ -110,6 +110,15 @@ func (i *Client) ListPods() ([]v1.Pod, error) {
 	return podEntries, nil
 }
 
+func (i *Client) GetPod(namespace, name string) (*v1.Pod, error) {
+	pod, err := i.clientSet.CoreV1().Pods(namespace).Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	return pod, nil
```
maiqueb, line 119: hm, maybe I'm missing something, but why don't you just `return i.clientSet.CoreV1().Pods(namespace).Get(context.TODO(), name, metav1.GetOptions{})` ?

maiqueb (commented):
Thanks for this patch; while it seems I'm making plenty of comments, I'm very happy with the quality of the code.

I wonder about the impacts of this retry on the code - especially since the retries are nearly immediate (no backoff is used). 

I honestly don't expect this scenario to happen (famous last words ...), but if the cronjob is triggered and there are plenty of pending pods (and staying in pending state for long), we will overload the API.

pkg/reconciler/iploop.go
xagent003: I am hesitant to use a backoff because remember that for #182 when the IP Pool is full, we WANT it to reconcile. In fact, we dont need any retries in that case... because the container will never come up otherwise. It is in Pending state, does not have an IP annotation either, and will match the podRef. 

So adding more retries or backoff would just add load or delay the reconciler (and thus blocked Pod) longer.

Any suggestions on how to distinguish between these two scenarios? Or should we just retry once

xagent003:
@dougbtv @maiqueb Can you please take a look, revised the PR. The test failure looks unrelated, and same as my PR https://github.com/k8snetworkplumbingwg/whereabouts/pull/180 is hitting.

Can you re-run/override?

maiqueb:
> @dougbtv @maiqueb Can you please take a look, revised the PR. The test failure looks unrelated, and same as my PR #180 is hitting.
> 
> Can you re-run/override?

Would you rebase both #211 (which fixes the unit test issue you're facing) ? We cannot override, since this is not executing any unit tests.

pkg/reconciler/iploop.go
```diff
@@ -102,19 +103,73 @@ func (rl *ReconcileLooper) findOrphanedIPsPerPool(ipPools []storage.IPPool) erro
...
+				 * so re-fetch the Pod 5x
+				 */
+				podToMatch := &livePod
+				retries := 0
+
+				logging.Debugf("Re-fetching Pending Pod: %s IP-to-match: %s", livePodRef, ip)
+
+				for retries < storage.PodRefreshRetries {
+					retries += 1
+					podToMatch = rl.refreshPod(livePodRef)
```
maiqueb, line 119: would it make sense to instead of retrying now just put the pods in pending state to a list, and after the main loop (that handles all pods that are **not** pending) iterate all the ones that were in pending state ? 

I am unsure if I am overthinking this.

PR #196, opened 2022-02-21
Simple go e2e test

nicklesimba:
This PR adds a simple golang e2e test and provides a framework for future e2e tests, as well as some vendoring changes to support the framework, specifically for net-attach-def client.

The test added here creates a net-attach-def for whereabouts and a single pod with an annotation that uses whereabouts config. It also adds a "pod-status.go" file with test functions for checking the status of a pod.


nicklesimba:
This PR has been deprecated by #197

PR #185, opened 2022-02-04
IP control loop

maiqueb:
Introduce an IP reconciliation loop for whereabouts. 

The controller monitor pod deletion events, and for every deleted pod it:
  - reads the pod's network status annotations
  - read the network attachment for every network referenced in the status
  - extract the associated IP range of the attachment
  - read the corresponding IP pool
  - delete all existing attachments on that pool for the pod being deleted

The controller is unit tested, and e2e tests are introduced to check for stale IP addresses for both pods and stateful sets.

It is updating the allocations in the datastore via the whereabouts API, which uses leader election to allow concurrent updates to the IP pools.

It only reads the `Pod`s / `IPPool`s / `Net-Attach-Def`s from the informer cache, in order to minimize the load of the Kubernetes API.

Depends-on: #202 
Fixes: #182

PR #197, opened 2022-02-22
Provide minimal test for whereabouts, along with successful vendor of the net-attach-def client for k8s 1.22

maiqueb:
Worked on top of PR #196.


e2e/e2e_test.go
```diff
@@ -229,24 +248,31 @@ func filterNetworkStatus(networkStatuses []nettypes.NetworkStatus, predicate fun
...
+		return "", fmt.Errorf("the pod does not have IPs for its secondary interfaces")
+	}
+
+	return secondaryInterfaceNetworkStatus.IPs[0], nil
 }
 
 func inRange(cidr string, ip string) bool {
 	_, cidrRange, err := net.ParseCIDR(cidr)
-	Expect(err).To(BeNil())
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
```
nicklesimba, line 275: I'm in favor of making inRange return `(bool, error)` rather than using this ExpectWithOffset statement - just to be consistent with removing Expect statements in helper functions. Like so:

```
func inRange(cidr string, ip string) bool, error {
	_, cidrRange, err := net.ParseCIDR(cidr)
	if err != nil {
		return false, err
	}
	ipInRangeCandidate := net.ParseIP(ip)

	return cidrRange.Contains(ipInRangeCandidate)
}
```

And earlier in the code, the following can be changed to accomodate this
```
Expect(inRange(ipv4TestRange, secondaryIfaceIP)).To(BeTrue())

becomes...

ipInRange, err = inRange(ipv4TestRange, secondaryIfaceIP)
Expect(ipInRange).To(BeTrue())
Expect(err).To(BeNil())
```

e2e/e2e_test.go
```diff
@@ -3,89 +3,61 @@ package whereabouts_e2e
...
-			label = make(map[string]string)
-			annotations = make(map[string]string)
-			annotations["k8s.v1.cni.cncf.io/networks"] = testNetworkName
+			netAttachDef = macvlanNetworkWithWhereaboutsIPAMNetwork()
 
-			// Create a net-attach-def
 			By("creating a NetworkAttachmentDefinition for whereabouts")
 			_, err = clientInfo.addNetAttachDef(netAttachDef)
-			Expect(err).To(BeNil())
+			Expect(err).NotTo(HaveOccurred())
```
nicklesimba, line 60: Why change from `To(BeNil())` to `NotTo(HaveOccurred())` ? 
`To(BeNil())` reads more clearly without using negatives. Very pedantic comment, but I used `To(BeNil())` initially on your recommendation, so this is leaving me confused. Ditto on all other occurrences in this PR.

nicklesimba (commented):
I added a few questions and also changes I'd like to see. Feedback on my feedback welcome :)

e2e/e2e_test.go
maiqueb: I am just trying to avoid returning `(bool, error)` . That return type makes no sense to me; it should be re-written instead to return an error - and the absence of one, means it worked. 

Your thoughts @nicklesimba ?

e2e/e2e_test.go
maiqueb: erm, yeah, if I said that, I was wrong. My bad then.

You should error check with `Expect(err).To(Succeed())` or `Expect(err).To(HaveOccurred())`:
https://github.com/onsi/gomega/blob/7e687ad94511b02c75ab34d08e02c5b1a68358b3/matchers.go#L53
> //HaveOccurred succeeds if actual is a non-nil error
//The typical Go error checking pattern looks like:
//    err := SomethingThatMightFail()
//    Expect(err).ShouldNot(HaveOccurred())

One thing is a non-nil thing, another a non-nil error. Same thing with `Succeed()` which is a few lines below.

e2e/e2e_test.go
nicklesimba: Your approach certainly reads much better. If removing "Expect"-ations from helper functions is a guideline rather than a strict requirement, then I'm okay with your current implementation.

e2e/e2e_test.go
nicklesimba: Possibly me being confused here as well, no worries. Thanks for the explanation

e2e/e2e_test.go
maiqueb: For completeness, my suggestion is to re-write this function to:

```golang
func inRange(cidr string, ip string) error {
	_, cidrRange, err := net.ParseCIDR(cidr)
	if err != nil {
		return err
	}
	if cidrRange.Contains(net.ParseIP(ip)) {
            return nil
        }
        return fmt.Errorf("ip [%s] is NOT in range %s", ip, cidr)
}
```

Which will be used in the tests like: 
```golang
// success
Expect(inRange(ipv4TestRange, secondaryIfaceIP)).To(Succeed())

// failure
Expect(inRange(ipv4TestRange, secondaryIfaceIP)).To(HaveOccurred())
```

maiqueb: > Your approach certainly reads much better. If removing "Expect"-ations from helper functions is a guideline rather than a strict requirement, then I'm okay with your current implementation.

Yes, it is not a strict requirement. I just feel expectations belong in the test body. 

For instance, the `ExpectWithOffset` exists exactly to improve the readability of expectations on helpers.

maiqueb: Done.

nicklesimba (commented):
LGTM, @dougbtv can you give this a pass when you get a chance as well?

pkg/config/config.go
```diff
@@ -241,10 +244,8 @@ func configureStatic(n *types.Net, args types.IPAMEnvArgs) error {
 
 	// CNI spec 0.2.0 and below supported only one v4 and v6 address
 	if numV4 > 1 || numV6 > 1 {
-		for _, v := range types020.SupportedVersions {
-			if n.CNIVersion == v {
-				return fmt.Errorf("CNI version %v does not support more than 1 address per family", n.CNIVersion)
-			}
+		if isSupportedVersion, _ := version.GreaterThanOrEqualTo(n.CNIVersion, minimumSupportedVersion); !isSupportedVersion {
```
dougbtv, line 247: Much nicer!

e2e/pod_status.go
```diff
@@ -0,0 +1,93 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+	return podList, nil
+}
+
+// WaitForPodBySelector waits up to timeout seconds for all pods in 'namespace' with given 'selector' to enter provided state
+// If no pods are found, return nil.
+func WaitForPodBySelector(cs *kubernetes.Clientset, namespace, selector string, timeout int) error {
```
dougbtv, line 77: Do we call this anywhere? We don't have to remove it, totally ok to have, just making sure I'm reading it correctly.

dougbtv (approved):
This is a solid LGTM. Seems like you guys got this squeaky clean before the submission.

e2e/pod_status.go
nicklesimba: Good catch! This was left in as part of the more complex e2e test that I'm working on. We can keep it in if it's not a big deal since it will get used eventually, but I'm okay with removing it, too.

dougbtv: No prob at all. Let's keep it!

maiqueb:
/hold

@dougbtv I'm having some doubts about the proposed vendoring here - for whatever reason I am not being able to get this to work with #185. Give me a while to figure this one out... 

I'll also rebase the PR with latest master.

PR #199, opened 2022-02-24
vendoring: properly vendor controller-gen

maiqueb:
This patch should address the constast hassle exhibited whenever
a user has to run `go mod tidy && go mod vendor` using go <= 1.16.

With this, we run the vendored binary.


maiqueb:
Abandoning in lieu of #200 , which achieves the same, but removes the dependency instead of vendoring it.

PR #203, opened 2022-02-28
Remove golint

maiqueb:
Replace golint - which is deprecated - with the combination of `go vet` + `staticheck`, 
which is recommended by the `golint` project in their [readme](https://github.com/golang/lint#readme).


dougbtv (approved):
LGTM! Thanks for the eye on go lint being obsolete.

PR #204, opened 2022-03-01
build, dockerfiles: add reconciler

maiqueb:
The ip-reconciler is only currently being shipped in the amd64 images.
This commit adds it to both the arm64 and openshift images, thus
aligning all of them.


dougbtv (approved):
Yep! Good catch. LGTM

PR #206, opened 2022-03-08
E2e tests check stateful sets

maiqueb:


maiqueb:
Hi @xagent003 . I'm trying to come up with e2e test to reproduce the issues you've reported about the leaking IP addresses. 

Would you take a look at the top 2 commits in this PR to see if it makes sense to you ? 

I have unfortunately not been able to reproduce them so far; would appreciate your help to figure out what I'm doing wrong.

e2e/e2e_test.go
```diff
@@ -87,6 +92,85 @@ var _ = Describe("Whereabouts functionality", func() {
...
+				_, err = clientInfo.provisionStatefulSet(statefulSetName, namespace, serviceName, initialReplicaNumber)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(
+					clientInfo.Client.CoreV1().Pods(namespace).List(
+						context.TODO(), metav1.ListOptions{LabelSelector: selector})).To(
+					WithTransform(func(podList *core.PodList) int { return len(podList.Items) }, Equal(initialReplicaNumber)))
+			})
+
+			AfterEach(func() {
+				const noReplicas = 0
```
nicklesimba: This is more feedback for myself...that I want to run by you. I like this tight scoping for consts, i.e. allow const to exist at the minimal possible scope. I think I'm going to follow this for my comprehensive e2e PR.

Example: `singlePodName   = "whereabouts-basic-test"` declared somewhere around line 30-40 should be removed from being a global const, and should instead live under the `BeforeEach()` of "Single pod tests" context, since it's only ever used there.

And if a const ends up being reused its scope can be changed as needed.

e2e/pod_status.go
```diff
@@ -49,6 +50,45 @@ func isPodGone(cs *kubernetes.Clientset, podName, namespace string) wait.Conditi
...
+		if err != nil && !k8serrors.IsNotFound(err) {
+			return false, fmt.Errorf("something weird happened with the stateful set whose status is: [%s]. Errors: %w", statefulSet.Status.String(), err)
+		}
+
+		associatedPods, err := cs.CoreV1().Pods(namespace).List(context.TODO(), selectViaLabels(labelSelector))
+		if err != nil {
+			return false, err
+		}
+
+		return isStatefulSetEmpty(statefulSet) && areAssociatedPodsGone(associatedPods), nil
```
nicklesimba, line 76: Garbage collection (foreground cascading deletion) won't work for cleaning up pods, I guess?

e2e/e2e_test.go
maiqueb: Yes, that's the idea. 

If the value of the const happens to be the same ... It might make sense to move it to a common place, but, that's not a silver bullet.

e2e/pod_status.go
maiqueb: It does, I just haven't used it yet.

e2e/e2e_test.go
```diff
@@ -330,23 +342,18 @@ func generateNetAttachDefSpec(name, namespace, config string) *nettypes.NetworkA
...
+		"mode": "bridge",
+		"ipam": {
+			"type": "whereabouts",
+			"leader_lease_duration": 1500,
+			"leader_renew_deadline": 1000,
+			"leader_retry_period": 500,
+			"range": "10.10.0.0/16",
+			"log_level": "debug",
+			"log_file": "/tmp/wb"
+		}
```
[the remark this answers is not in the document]
maiqueb: https://github.com/k8snetworkplumbingwg/whereabouts/pull/185/commits/4f2a6911270b850e02f356bbf17bc9543ba15285

I was wrong in adapting this in this PR, it is not needed, since whereabouts is only receiving the configuration for a single plugin at a time. On the other hand, it is required for the ip reconciliation loop.

If you try the test I've added in that commit without the associated code changes, you'll see it fails.

maiqueb:
Abandoning since this change is featured in (already merged PR #185 ).

Issue #198, opened 2022-02-23
Change logging and history for ip-reconciler


xagent003:
- Change default log-level to debug. The error logging is not useful, all the issue so far have been in non-error conditions. At least for now until the ip-reconciler issues are all fixed and it is stable

 - Change back successfulJobsHistoryLimit to 3 or greater

 - Write to file/volume on host, allow configuring the log file location. The containers get cleaned up, then there is no way to debug why ip-reconciler cleaned up wrong IP or did not clean up an IP. It should log in a file on the host, and placed along with the whereabouts.log

PR #210, opened 2022-03-29
Use zap logger with logrotation, file logging for ip-reconciler

xagent003:
This solves many logging issues we've had with whereabouts and the ip-reconciler. We already have this working on our custom whereabouts fork we use in production (https://github.com/platform9/whereabouts), would like to get it merged upstream:

1. ip-reconciler logs are lost once the Pod is terminated. This was made worse by changing the succesful job history to 0 since all the bugs we've seen were when the Job was Succesful. So, have it mount a volume on host and write to a proper log file. This way if container is cleaned up, the logs are still accessible.

2. whereabouts is using a custom logging package which is a wrapper around Printf. Instead, use a common logger like Zap logger. Also, use lumberjack package for built in file rotation. This works for both whereabouts and new file based ip-reconciler logging in #1, and the logs are rotated automatically.

Yes, the default logging level is debug. I think we need it until this matures as we see a lot of whereabouts and ip-reconciler issues that simply would not be debuggable otherwise.


cmd/reconciler/ip.go
```diff
@@ -13,11 +13,11 @@ const defaultReconcilerTimeout = 30
 
 func main() {
 	kubeConfigFile := flag.String("kubeconfig", "", "the path to the Kubernetes configuration file")
-	logLevel := flag.String("log-level", "error", "the logging level for the `ip-reconciler` app. Valid values are: \"debug\", \"verbose\", \"error\", and \"panic\".")
+	logFile := flag.String("log-file", "/host/var/log/ip-reconciler.log", "File on host for ip-reconciler to log to")
```
maiqueb, line 16: I don't understand why you are removing the log level parameter. It still makes sense.

Also, you cannot use this default - i.e. what if the `cronjob` spec does not feature the volume mount ? It will fail.

You should use a default on the container file system, but override that in the daemonset spec shipped in whereabouts.

maiqueb:
> Fix for #198
> 
> This solves many logging issues we've had with whereabouts and the ip-reconciler. We already have this working on our custom whereabouts fork we use in production (https://github.com/platform9/whereabouts), would like to get it merged upstream:
> 
> 1. ip-reconciler logs are lost once the Pod is terminated. This was made worse by changing the succesful job history to 0 since all the bugs we've seen were when the Job was Succesful. So, have it mount a volume on host and write to a proper log file. This way if container is cleaned up, the logs are still accessible.

Makes sense to me. 

> 2. whereabouts is using a custom logging package which is a wrapper around Printf. Instead, use a common logger like Zap logger. Also, use lumberjack package for built in file rotation. This works for both whereabouts and new file based ip-reconciler logging in point 1 above, and the logs are rotated automatically.

Makes sense to me. Thanks for the effort in this one.

> 
> Yes, the default logging level is debug. I think we need it until this matures as we see a lot of whereabouts and ip-reconciler issues that simply would not be debuggable otherwise.

I don't understand why you are removing the `log-level` configuration parameter. It still makes sense to have it.

doc/crds/ip-reconciler-job.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     app: whereabouts
 spec:
   concurrencyPolicy: Forbid
-  successfulJobsHistoryLimit: 0
+  successfulJobsHistoryLimit: 3
```
maiqueb, line 11: If you're now logging to the host's file-system, why would we want to preserve these around ?

maiqueb (changes requested):
Shouldn't the vendored files be added as well ? I think you haven't `go mod vendor` - but I might be wrong.

xagent003:
@maiqueb you're right, I think I can make the log-level configureable in zap/lumberjack... will try adding that.

PR #218, opened 2022-05-05
[DoNotMerge] Ip reconciler alerting

nicklesimba:
WIP pull request for ip-reconciler alerting. This PR has two goals:

1. enable the ip-reconciler to be invoked via the ip-control-loop, deprecating the ip-reconciler binary.
2. send metrics to prometheus when the ip-reconciler succeeds, so that cluster-network-operator can read the metrics and send alerts accordingly. (as a result, whereabouts code will interact with CNO yaml)


maiqueb:
First and foremost, if this PR has 2 objectives, I would rather have 2 PRs with one objective each.

That will help us focus, and enable you to progress further faster.

cmd/reconciler/ip.go
```diff
@@ -1,40 +1,46 @@
...
-	reconcilerTimeout := flag.Int("timeout", defaultReconcilerTimeout, "the value for a request timeout in seconds.")
-	flag.Parse()
+func InvokeIPReconciler(returnErr chan error) {
+	go func() {
+		doInvokeIPReconciler(returnErr)
+	}()
+	select {
+	case <-time.After(5 * time.Second):
+		returnErr <- errors.New("timed out (this is the 5 sec test timeout, nbd)")
+		return
```
maiqueb, line 23: you don't need to do this. You should pass a `context.WithTimeout` below, in line https://github.com/k8snetworkplumbingwg/whereabouts/pull/218/files#diff-99080d2a71bc34fba51f6b2bb1766239380ae51bdf70aa1496fa73b5ffdbbe79R31

I also suspect that the issue you're facing comes from spawning 2 go routines
 - first in https://github.com/k8snetworkplumbingwg/whereabouts/pull/218/files#diff-dbddff16de1ff73a326e91a361beb31cae41142fa4b9e85d5e759575af2a56bdR98
 - second in https://github.com/k8snetworkplumbingwg/whereabouts/pull/218/files#diff-99080d2a71bc34fba51f6b2bb1766239380ae51bdf70aa1496fa73b5ffdbbe79R17

but I am not 100% sure of this second claim.

PR #220, opened 2022-05-10
Refactor ippool checker into struct

maiqueb:
This PR moves the IPPool checks to a dedicated pkg.
    
It furthermore redesigns the pool consistency checks to use a struct, thus allowing
developers to write tests in the following way:

```go
Expect(
    poolconsistency.NewPoolConsistencyCheck(
        ipPool, podList.Items).MissingIPs()).To(BeEmpty())
Expect(
    poolconsistency.NewPoolConsistencyCheck(
        ipPool, podList.Items).StaleIPs()).To(BeEmpty())
```
instead of what is used currently:
```go
 Expect(
    iPPoolConsistency(
        k8sIPAM,
        clientInfo.Client,
        testNamespace,
        ipPoolName,
        entities.ReplicaSetQuery(rsName))).To(Succeed())
```

The implementation above does offer more information on failures, and is
more ginkgo native than the current approach.

### DIFF
--- a/e2e/client/ippool.go
+++ b/e2e/client/ippool.go
@@ -1,7 +1,7 @@
 // This code is based on code from the following repository
 // https://github.com/bcreane/k8sutils
 
-package whereabouts_e2e
+package client
 
 import (
 	"context"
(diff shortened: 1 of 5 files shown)


nicklesimba (approved):
This looks like great changes. Thanks also for including tests for poolconsistency!

/lgtm

PR #221, opened 2022-05-11
storage, k8s: Fallback to kube-system on empty ns

qinqon:
When the kubeconfig has no namespace on the current context the request
to handle leader election resources fail with missing namespace since a
request like "/foobar" cannot be send. This change use "kube-system" as
the fallback one if the namespace is empty.

### DIFF
--- a/pkg/storage/kubernetes/ipam.go
+++ b/pkg/storage/kubernetes/ipam.go
@@ -19,6 +19,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/k8snetworkplumbingwg/whereabouts/pkg/allocate"
@@ -34,7 +35,7 @@ func NewKubernetesIPAM(containerID string, ipamConf whereaboutstypes.IPAMConfig)
 	if cfg, err := clientcmd.LoadFromFile(ipamConf.Kubernetes.KubeConfigPath); err != nil {
 		return nil, err
 	} else if ctx, ok := cfg.Contexts[cfg.CurrentContext]; ok && ctx != nil {
-		namespace = ctx.Namespace
+		namespace = wbNamespaceFromCtx(ctx)
 	} else {
 		return nil, fmt.Errorf("k8s config: namespace not present in context")
 	}
@@ -542,4 +543,12 @@ RETRYLOOP:
 	}
 
 	return newip, err
+}
+
+func wbNamespaceFromCtx(ctx *clientcmdapi.Context) string {
+	namespace := ctx.Namespace
+	if namespace == "" {
+		return metav1.NamespaceSystem
+	}
+	return namespace
 }


pkg/storage/kubernetes/ipam.go
```diff
@@ -35,6 +35,9 @@ func NewKubernetesIPAM(containerID string, ipamConf whereaboutstypes.IPAMConfig)
 		return nil, err
 	} else if ctx, ok := cfg.Contexts[cfg.CurrentContext]; ok && ctx != nil {
 		namespace = ctx.Namespace
+		if namespace == "" {
+			namespace = "kube-system"
```
maiqueb, line 39: I think you should use the `metav1.NamespaceSystem` constant instead of this hard-coded string. 

I would say I also miss a unit test for this, but the whole pkg as none .... Not good.

maiqueb (changes requested):
Having a fall-back / default namespace is a good call.

Never the less, I think you should use a constant instead of an hard-coded string.

Furthermore, I have a question: how can the kubeconfig have no namespace on the current context ? I am 100% sure I have been there, but I don't remember right now how I reached that situation.

maiqueb (approved):
Thanks.

pkg/storage/kubernetes/ipam.go
```diff
@@ -35,6 +35,9 @@ func NewKubernetesIPAM(containerID string, ipamConf whereaboutstypes.IPAMConfig)
 		return nil, err
 	} else if ctx, ok := cfg.Contexts[cfg.CurrentContext]; ok && ctx != nil {
 		namespace = ctx.Namespace
+		if namespace == "" {
+			namespace = metav1.NamespaceSystem
+		}
```
maiqueb, line 40: actually , would you mind exporting this into a function ? 

Something like

```golang
func wbNamespace(ctx *clientcmdapi.Context) string {
	namespace := ctx.Namespace
	if namespace == "" {
		return metav1.NamespaceSystem
	}
	return namespace
}
```

maiqueb: I think that's wrong. There are 2 deployment paths for whereabouts: 
- a simple CNI plugin - which relies on having a kubeconfig path - i.e. it doesn't have an in cluster config
- a controller / cronjob running in a pod, which rely on the in cluster config.

The above are exclusive: the developer should be explicit and know which to call. By doing this change, you are imo mixing these two types of clients, thus making it more confusing, and hard to reason about.

pkg/storage/kubernetes/ipam.go
```diff
@@ -31,17 +32,27 @@ import (
...
-	} else if ctx, ok := cfg.Contexts[cfg.CurrentContext]; ok && ctx != nil {
-		namespace = ctx.Namespace
+	var kubernetesClient *Client
+	if ipamConf.Kubernetes.KubeConfigPath == "" {
+		var err error
+		kubernetesClient, err = NewClient(storage.RequestTimeout)
+		if err != nil {
+			return nil, fmt.Errorf("failed instantiating kubernetes client: %v", err)
+		}
+		namespace = metav1.NamespaceSystem
```
maiqueb, line 42: this is wrong imo. See https://github.com/k8snetworkplumbingwg/whereabouts/pull/221#discussion_r870483371 .

maiqueb (changes requested):

pkg/storage/kubernetes/ipam.go
```diff
@@ -31,15 +32,16 @@ import (
...
 	} else if ctx, ok := cfg.Contexts[cfg.CurrentContext]; ok && ctx != nil {
-		namespace = ctx.Namespace
+		namespace = wbNamespaceFromCtx(ctx)
 	} else {
 		return nil, fmt.Errorf("k8s config: namespace not present in context")
 	}
-
-	kubernetesClient, err := NewClientViaKubeconfig(ipamConf.Kubernetes.KubeConfigPath, storage.RequestTimeout)
+	var err error
+	kubernetesClient, err = NewClientViaKubeconfig(ipamConf.Kubernetes.KubeConfigPath, storage.RequestTimeout)
```
maiqueb, line 44: Why is this better - and even related to this PR ?

pkg/storage/kubernetes/ipam.go
```diff
@@ -19,6 +19,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
```
maiqueb, line 22: why is this needed ?

maiqueb (commented):
One quick question: I think this might be a leftover.

pkg/storage/kubernetes/ipam.go
maiqueb: oh. Just noticed https://github.com/k8snetworkplumbingwg/whereabouts/pull/221/files#diff-7a77fc45b536e504372eba572994f95f478aa250d4cdb936bdc3d1018c86b09eR548 

My bad.

maiqueb (approved):
tks

dougbtv (approved):

PR #222, opened 2022-05-17
reconciler: account for pods that do not have net-status annotations

maiqueb:


rthakur-est (approved):
LGTM

PR #223, opened 2022-05-19
Bump golang1.17

maiqueb:
Golang 1.16 is EOL, as shown in the [golang supported versions doc](https://endoflife.date/go).

Fixes: #224

PR #232, opened 2022-05-25
build, golang1.17: sloppy ip parsing

maiqueb:
Adds backwards compatibility to existing pools by providing a custom unmarshaller of the `IPAMConfig` struct.

Depends-on: #223


maiqueb:
/hold 

Let me try another approach first - trying to write custom marshall / unmarshall functions for the `IPAMConfig` type.

maiqueb:
@dougbtv ready to review.

I'd appreciate if we kept things separate and merged #223 first. I would then rebase this one, which would show a lot less differences.

Finally, I think releasing a whereabouts version **before** bumping golang & kubernetes versions is a good move.

dougbtv (approved):
LGTM

PR #233, opened 2022-05-30
[DoNotMerge] Ip Reconciler Alerting

nicklesimba:
Preface: this is a dupe of #218 created to better separate changes into commits for easier code review. 

WIP pull request for ip-reconciler alerting. This PR introduces 2 changes:

1. enables the ip-reconciler to be invoked via the ip-control-loop, deprecating the ip-reconciler binary. This is done simply be reorganizing the files and removing a line in the build script that creates the reconciler binary.
2. sends metrics to prometheus when the ip-reconciler succeeds, so that cluster-network-operator can read the metrics and send alerts accordingly. (as a result, whereabouts code will interact with CNO yaml when everything is working correctly)

Aside from simple code cleanup, this is mostly a finished PR. Once functionality with alerting is confirmed to work with CNO, the code here in whereabouts should be ready to go.

Note: These changes are listed under the same pull request because it doesn't make sense to merge one without the other.


maiqueb:
First of, I don't understand why you abandoned #218 and started off another PR. The history of the reviews will be gone unless someone is really into digging through the chain of PRs you've created. 

Secondly, as I've stated in https://github.com/k8snetworkplumbingwg/whereabouts/pull/218#issuecomment-1122601857, this PR is doing 2 different things. As such, there should be 2 different PRs: 
- 1 for removing the reconciler binary, moving that logic into the main loop of IP reconciliation
- another for adding metrics

That allows the reviewers to focus on the problem at hand - and again, there are two.

I don't buy the argument that "it doesn't make sense to merge one without the other." It makes sense to merge one without the other. 

Do we want to get rid of the cronjob ? Yes. Why should we hold on that until we get the metrics right ? No logical reason I can think of. 

Furthermore, adding metrics introduces the prometheus dependencies which are adding +- 150 files to this repo. If everything related to metrics goes in a separate PR reviewers can focus on the refactor you're doing, thus increasing the odds the refactor is correctly introduced.

nicklesimba:
Closed in favor of #238 and #239

Issue #212, opened 2022-04-09
Dual Stack support


Ayush21298:
## Type

New feature or Enhancement of existing feature

## Description

Is there any plan for supporting Dual stack in near future ?
*As we all know that Kubernetes v1.20+ started supporting Dual stack and it is a stable feature now.*

Also, at present,  is there any method of assigning multiple IP addresses from different ranges ? *(I couldn't find one)*
If it exists, we can start developing Dual stack feature on top of it.

Ayush21298:
Thank you for your response @maiqueb :)
Sure, that will be a good start. Thank you for inviting me to the bi-monthly call.
Can you please let me know the schedule of the same ?
I could only find the info. about bi-weekly call [here](https://github.com/k8snetworkplumbingwg/community#meetings).

maiqueb:
> Thank you for your response @maiqueb :) Sure, that will be a good start. Thank you for inviting me to the bi-monthly call. Can you please let me know the schedule of the same ? I could only find the info. about bi-weekly call [here](https://github.com/k8snetworkplumbingwg/community#meetings).

Oh I just had my wording word and confused bi-monthly with bi-weekly. That's the one - once every 2 weeks.

hhhybb:
Sounds cool, and hope to see the dual stack support soon.

Ayush21298:
@dougbtv can you please assign this issue to me based on #250 ? :)

PR #237, opened 2022-06-08
DualStack Proposal

Ayush21298:
# Proposal

Brief Design Proposal for IPv4/IPv6 DualStack support.

## Checklist

- [x] Feature introduction
- [x] Development Goals
- [x] Design idea
- [x] Impact on modules
- [x] Alternative Design

_I wasn't sure about the proposal format being used for the project, so if anything is missing please add it in the checklist :)_

### Related Issue

[Dual Stack Support](https://github.com/k8snetworkplumbingwg/whereabouts/issues/212)

### DIFF
--- a/doc/proposals/dualstack.md
+++ b/doc/proposals/dualstack.md
@@ -1,0 +1,280 @@
+# Whereabouts support for Ipv4/IPv6 DualStack
+
+# Table of contents
+
+- [Introduction](#introduction)
+  - [Goal](#goal)
+- [Design](#design)
+  - [Changes in IPAM Config](#configChanges)
+  - [Changes in Modules](#moduleChanges)
+- [Alternative Design](#alternative)
+- [Summary](#summary)
+
+<hr>
+
+## Introduction <a name="introduction"></a>
+
+Starting v1.20, Kubernetes officially started supporting DualStack, which is being used extensively in many projects. 
+
+Whereabouts being an important IPAM CNI plugin, should also make a plan for supporting DualStack.
+This is a design proposal document for introducing IPv4/IPv6 DualStack support in whereabouts.
+
+
+### Goal <a name="goal"></a>
+
+- Allocate IPv4 and/or IPv6 addresses to pods depending on IPAM configuration
+- Above feature should not break the backward compatibility for SingleStack configuration
+- Allocate multiple (even more than 2) IPs to the pods irrespective of IPv4/IPv6 _(Optional and depends on design choice)_
+
+<hr>
+
+## Design <a name="design"></a>
+
+The idea is to update the IPAM config format to include multiple IP configurations.
+
+Basically, rather than having IP related configurations directly as key members of `IPAMConfig` object, they be included as element of an object type, let's say `IPConfig` and `IPAMConfig` should have an array of `IPConfig` as it's element.
+
+### Changes in IPAM Config <a name="configChanges"></a>
+
+This will lead to change in IPAM config something as follows:
+
+<table>
+<tr>
+<th>Old IPAM Config</th>
+<th>Changes</th>
+<th>New IPAM Config</th>
+</tr>
+<tr>
+<td>
+  
+```json
+{
+      "cniVersion": "0.3.0",
+      "name": "whereaboutsexample",
+      "type": "macvlan",
+      "master": "eth0",
+      "mode": "bridge",
+      "ipam": {
+        "type": "whereabouts",
+        "range": "192.168.2.225/28",
+        "exclude": [
+           "192.168.2.229/30",
+           "192.168.2.236/32"
+        ]
+      }
+}
+```
+  
+</td>
+<td>
+
+```diff
+{
+      "cniVersion": "0.3.0",
+      "name": "whereaboutsexample",
+      "type": "macvlan",
+      "master": "eth0",
+      "mode": "bridge",
+      "ipam": {
+        "type": "whereabouts",
+-       "range": "192.168.2.225/28",
+-       "exclude": [
+-          "192.168.2.229/30",
+-          "192.168.2.236/32"
+-       ]
++       "ips": [
++         {
++           "range": "192.168.2.225/28",
++           "exclude": [
++              "192.168.2.229/30",
++              "192.168.2.236/32"
++           ]
++         },
++         {
++           "range": "2001::0/116",
++         }
++       ]
+      }
+}
+```
+
+</td>
+<td>
+
+```json
+{
+      "cniVersion": "0.3.0",
+      "name": "whereaboutsexample",
+      "type": "macvlan",
+      "master": "eth0",
+      "mode": "bridge",
+      "ipam": {
+        "type": "whereabouts",
+        "ips": [
+          {
+            "range": "192.168.2.225/28",
+            "exclude": [
+               "192.168.2.229/30",
+               "192.168.2.236/32"
+            ]
+          },
+          {
+            "range": "2001::0/116",
+          }
+        ]
+      }
+}
+```
+
+</td>
+</tr>
+</table>
+
+_Note: But at the same time we also need to support the old field (i.e. field duplication) in order to make the design backward compatible._
+
+### Changes in Modules <a name="moduleChanges"></a>
+
+#### whereabouts/pkg/types/types.go
+
+```diff
+
++type IPConfig struct {
++     Addresses           []Address         `json:"addresses,omitempty"`
++     OmitRanges          []string          `json:"exclude,omitempty"`
++     Range               string            `json:"range"`
++     RangeStart          net.IP            `json:"range_start,omitempty"`
++     RangeEnd            net.IP            `json:"range_end,omitempty"`
++}
+
+ type IPAMConfig struct {
+      Name                string
+      Type                string            `json:"type"`
+      Routes              []*cnitypes.Route `json:"routes"`
+      Datastore           string            `json:"datastore"`
++     IPs                 []IPConfig        `json:"ips"`
+      Addresses           []Address         `json:"addresses,omitempty"`
+      OmitRanges          []string          `json:"exclude,omitempty"`
+      DNS                 cnitypes.DNS      `json:"dns"`
+      Range               string            `json:"range"`
+      RangeStart          net.IP            `json:"range_start,omitempty"`
+      RangeEnd            net.IP            `json:"range_end,omitempty"`
+      GatewayStr          string            `json:"gateway"`
+      EtcdHost            string            `json:"etcd_host,omitempty"`
+      EtcdUsername        string            `json:"etcd_username,omitempty"`
+      EtcdPassword        string            `json:"etcd_password,omitempty"`
+      EtcdKeyFile         string            `json:"etcd_key_file,omitempty"`
+      EtcdCertFile        string            `json:"etcd_cert_file,omitempty"`
+      EtcdCACertFile      string            `json:"etcd_ca_cert_file,omitempty"`
+      LeaderLeaseDuration int               `json:"leader_lease_duration,omitempty"`
+      LeaderRenewDeadline int               `json:"leader_renew_deadline,omitempty"`
+      LeaderRetryPeriod   int               `json:"leader_retry_period,omitempty"`
+      LogFile             string            `json:"log_file"`
+      LogLevel            string            `json:"log_level"`
+      OverlappingRanges   bool              `json:"enable_overlapping_ranges,omitempty"`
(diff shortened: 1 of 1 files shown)


doc/proposals/dualstack.md
```diff
@@ -0,0 +1,280 @@
...
+```diff
+{
+      "cniVersion": "0.3.0",
+      "name": "whereaboutsexample",
+      "type": "macvlan",
+      "master": "eth0",
+      "mode": "bridge",
+      "ipam": {
+        "type": "whereabouts",
+-       "range": "192.168.2.225/28",
```
dougbtv, line 80: I think we both want to keep this for backwards compatibility.

dougbtv: Ooops, I see this is one example, I see that other keeps this.

Ayush21298: Yes :)

doc/proposals/dualstack.md
```diff
@@ -0,0 +1,280 @@
...
+
+<hr>
+
+## Introduction <a name="introduction"></a>
+
+Starting v1.20, Kubernetes officially started supporting DualStack, which is being used extensively in many projects. 
+
+Whereabouts being an important IPAM CNI plugin, should also make a plan for supporting DualStack.
+This is a design proposal document for introducing IPv4/IPv6 DualStack support in whereabouts.
+
```
maiqueb, line 21: nit: remove this extra new line.

doc/proposals/dualstack.md
```diff
@@ -0,0 +1,280 @@
...
+  - [Goal](#goal)
+- [Design](#design)
+  - [Changes in IPAM Config](#configChanges)
+  - [Changes in Modules](#moduleChanges)
+- [Alternative Design](#alternative)
+- [Summary](#summary)
+
+<hr>
+
+## Introduction <a name="introduction"></a>
```
maiqueb, line 15: Isn't this `<a name="introduction"></a> ` redundant in markdown ? AFAIU, there's an implicit anchor after the lowercase, dash separated, name of the heading. In this case - introduction.

doc/proposals/dualstack.md
```diff
@@ -0,0 +1,280 @@
...
+
+Starting v1.20, Kubernetes officially started supporting DualStack, which is being used extensively in many projects. 
+
+Whereabouts being an important IPAM CNI plugin, should also make a plan for supporting DualStack.
+This is a design proposal document for introducing IPv4/IPv6 DualStack support in whereabouts.
+
+
+### Goal <a name="goal"></a>
```
maiqueb, line 23: same; `goal` can be used directly without specifying the anchor name.

doc/proposals/dualstack.md
```diff
@@ -0,0 +1,280 @@
...
+
+- Allocate IPv4 and/or IPv6 addresses to pods depending on IPAM configuration
+- Above feature should not break the backward compatibility for SingleStack configuration
+- Allocate multiple (even more than 2) IPs to the pods irrespective of IPv4/IPv6 _(Optional and depends on design choice)_
+
+<hr>
+
+## Design <a name="design"></a>
```
maiqueb, line 31: ditto

doc/proposals/dualstack.md
```diff
@@ -0,0 +1,280 @@
...
+
+The idea is to update the IPAM config format to include multiple IP configurations.
+
+Basically, rather than having IP related configurations directly as key members of `IPAMConfig` object, they be included as element of an object type, let's say `IPConfig` and `IPAMConfig` should have an array of `IPConfig` as it's element.
+
+### Changes in IPAM Config <a name="configChanges"></a>
```
maiqueb, line 37: here it makes sense - assuming you don't want to use the implicit `changes-in-ipam-config` anchor name.

doc/proposals/dualstack.md
```diff
@@ -0,0 +1,280 @@
...
+
+- Allocate IPv4 and/or IPv6 addresses to pods depending on IPAM configuration
+- Above feature should not break the backward compatibility for SingleStack configuration
+- Allocate multiple (even more than 2) IPs to the pods irrespective of IPv4/IPv6 _(Optional and depends on design choice)_
+
+<hr>
+
+## Design <a name="design"></a>
+
+The idea is to update the IPAM config format to include multiple IP configurations.
```
maiqueb, line 33:
```suggestion
The IPAM configuration format would be updated to include multiple IP configurations.
```

doc/proposals/dualstack.md
```diff
@@ -0,0 +1,280 @@
...
+
+Basically, rather than having IP related configurations directly as key members of `IPAMConfig` object, they be included as element of an object type, let's say `IPConfig` and `IPAMConfig` should have an array of `IPConfig` as it's element.
```
maiqueb, line 35:
```suggestion
Rather than having IP related configurations directly as key members of `IPAMConfig` object, the IP configuration would be wrapped in a new struct - let's say `IPConfig`.
The existing `IPAMConfig` would have an array of `IPConfig` as an attribute.
```

doc/proposals/dualstack.md
```diff
@@ -0,0 +1,280 @@
...
+
+### Changes in IPAM Config <a name="configChanges"></a>
+
+This will lead to change in IPAM config something as follows:
+
+<table>
```
maiqueb, line 41: Using a table was a nice decision.

doc/proposals/dualstack.md
```diff
@@ -0,0 +1,280 @@
...
+      "master": "eth0",
+      "mode": "bridge",
+      "ipam": {
+        "type": "whereabouts",
+        "range": "192.168.2.225/28",
++       "secondary_range": "2001::0/116",
+        "exclude": [
+           "192.168.2.229/30",
+           "192.168.2.236/32",
++          "2001::0/124"
```
maiqueb, line 237: how would you map the entries on the exclusion list to the correct range ? Here it actually seems doable since the ranges are from different IP families, but, what if you want to use this feature to configure multiple IPs in the same interface ?

doc/proposals/dualstack.md
```diff
@@ -0,0 +1,280 @@
...
+
+<hr>
+
+### Summary <a name="summary"></a>
+
+Currently, we have above two approaches for supporting DualStack in whereabouts.
+
+First approach requires addition on a new _type_ for encapsulating all the IP related information on `IPAMConfig` and requires significant amount of change in the current code. But with this additional effort, it will have additional benefit of allowing us to allocate as many as we required IP addresses of any IP family without any constraints. _(We need to support the existing fields too for backward compatibility)_
+
+Second approach is somewhat simpler. It just adds some additional fields for _secondary_ IP address related configuration (which might be IPv4 or IPv6 depending on stack policy). The downside is that we will limit ourselves to adding at max 2 IP address for a pod. _(TBH, it also seems okay for now, but I prefer the first approach)_
```
maiqueb, line 280: I agree with the analysis presented here: first approach is preferable - simply because the second one does **not** scale.

Furthermore, the IP is somewhat clunky IMO.

I vote for option 1 as well.

maiqueb (changes requested):
Thanks for this proposal.

I'd like to actually congratulate you on the focus on the changes from the current API, and how easy you made it for us to understand them. Kudos.

I am OK with pretty much everything you wrote, but still pointed a couple of nits.

doc/proposals/dualstack.md
```diff
@@ -0,0 +1,280 @@
...
++     RangeStart          net.IP            `json:"range_start,omitempty"`
++     RangeEnd            net.IP            `json:"range_end,omitempty"`
++}
+
+ type IPAMConfig struct {
+      Name                string
+      Type                string            `json:"type"`
+      Routes              []*cnitypes.Route `json:"routes"`
+      Datastore           string            `json:"datastore"`
++     IPs                 []IPConfig        `json:"ips"`
```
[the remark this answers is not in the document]
dougbtv: `RangeConfiguration` ++

doc/proposals/dualstack.md
Ayush21298: We can take separate unions of IPv4 and IPv6 excluded ranges and just skip the IPs that belongs to that union set/list :)
I don't think we need to map it to the exact range. ^^

maiqueb (changes requested):
Nearly there. 

Are you ready to propose a test plan for the feature, or would you prefer to do so later on ? 

I suggest you never the less familiarize yourself with the [e2e tests we have](https://github.com/k8snetworkplumbingwg/whereabouts/blob/master/e2e/e2e_test.go); imo, you would add a new test suite there.

Ayush21298:
Thank you for your comments.
I made the changes as suggested :)
 
> Are you ready to propose a test plan for the feature, or would you prefer to do so later on ?

I'm planning to do it latter along with the development :)

> I suggest you never the less familiarize yourself with the [e2e tests we have](https://github.com/k8snetworkplumbingwg/whereabouts/blob/master/e2e/e2e_test.go); imo, you would add a new test suite there.

Thank you for you guidance :) 
I had a look at the unit test suites as well as e2e test suite and will update them as I go on along with development :)

Ayush21298:
Hello @dougbtv  @maiqueb  ~

Can you please review this :sweat_smile:

maiqueb (changes requested):
I have a question about the proposed alternative: how do the new `RangesConfiguration` slice play along w/ the *old* `Range` attribute ? Is the behavior additive (i.e. 1 range + 2 entries in the `RangeConfiguration` results in 3 configured interfaces) ? 

I would like to see a sub-section about this, explaining (w/ a couple of examples) how it would works.

Some alternatives for this I see:
- additive behavior (described above)
- only use the new range configuration array / slice (the *old* `Range` would be just used for reading - i.e. copy the first range from the array there)
- if `Range` is provided, disallow multiple ranges

Ayush21298:
> I have a question about the proposed alternative: how do the new `RangesConfiguration` slice play along w/ the _old_ `Range` attribute ? Is the behavior additive (i.e. 1 range + 2 entries in the `RangeConfiguration` results in 3 configured interfaces) ?
> 
> I would like to see a sub-section about this, explaining (w/ a couple of examples) how it would works.
> 
> Some alternatives for this I see:
> 
>     * additive behavior (described above)
> 
>     * only use the new range configuration array / slice (the _old_ `Range` would be just used for reading - i.e. copy the first range from the array there)
> 
>     * if `Range` is provided, disallow multiple ranges

@maiqueb As you described, when both old configuration and `IPRanges` are present, the behavior will be additive.

I have added a section for the same along with examples in the proposal. :)

maiqueb (approved):
Thank you for providing the examples - your proposal is now - imo - crystal clear.

I'll let this soak for a few weeks to give the community the time to also provide feedback / @dougbtv to back it up, but, I think you can start implementing this as is.

Good work.

Ayush21298:
> Thank you for providing the examples - your proposal is now - imo - crystal clear.
> 
> I'll let this soak for a few weeks to give the community the time to also provide feedback / @dougbtv to back it up, but, I think you can start implementing this as is.
> 
> Good work.

Thanks @maiqueb :)
Yes I started working on the same line in PR #250. I'll try to finish it up ASAP :)

maiqueb:
> > Thank you for providing the examples - your proposal is now - imo - crystal clear.
> > I'll let this soak for a few weeks to give the community the time to also provide feedback / @dougbtv to back it up, but, I think you can start implementing this as is.
> > Good work.
> 
> Thanks @maiqueb :) Yes I started working on the same line in PR #250. I'll try to finish it up ASAP :)

Take your time, no rush.

Feel free to reach out if you need help with the e2e tests.

JanScheurich:
We fully support this initiative to let whereabouts assign IP addresses from multiple IP ranges (IPv4 and/or v6) to a secondary pod interface and agree with the proposed implementation.

Ayush21298:
@dougbtv Hi, can we merge this proposal ?

dougbtv (approved):
LGTM. Thanks Ayush for the proposal, looking forward to the functionality.

ryanlyy:
all,

which release will support this feature? seems now only master has these code support.

Thanks

dougbtv:
Thanks @ryanlyy -- we have published https://github.com/k8snetworkplumbingwg/whereabouts/releases/tag/v0.6 -- including this feature!

ryanlyy:
@dougbtv thank you!

lubronzhan:
Hi @Ayush21298 does this only support macvlan mode? looking at the test, only macvlan is tested.
Do you worried that with ipvlan, different pod might get same slaac ip through dhcp?
Thank you

PR #238, opened 2022-06-09
Moved IP Reconciler code into IP Control Loop

nicklesimba:
This commit removes the IP Reconciler as a separate binary, and also removes the cronjob that launched it. 

Instead, the code now invokes the reconciler code at a user configurable interval (once per day default). 

As part of these changes, the code has also been reorganized with more clear file names etc.


cmd/controlloop/controlloop.go
```diff
@@ -56,6 +61,26 @@ func main() {
 	}
 
 	networkController.Start(stopChan)
+
+	s := gocron.NewScheduler(time.UTC)
+	s.Cron("0 30 4 1/1 * ? *").Do(func() { // every day at 4:30 UTC (user configurable cron expression)
```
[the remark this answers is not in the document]
dougbtv: Also, we'll need some high-level documentation for it.

README.md
```diff
@@ -45,9 +45,9 @@ git clone https://github.com/k8snetworkplumbingwg/whereabouts && cd whereabouts
 kubectl apply \
     -f doc/crds/daemonset-install.yaml \
     -f doc/crds/whereabouts.cni.cncf.io_ippools.yaml \
-    -f doc/crds/whereabouts.cni.cncf.io_overlappingrangeipreservations.yaml \
-    -f doc/crds/ip-reconciler-job.yaml
+    -f doc/crds/whereabouts.cni.cncf.io_overlappingrangeipreservations.yaml
 ```
+^^^ i removed ip-reconciler-job.yaml. do we need to replace this with an ip-control-loop.yaml? I'm not familiar with this use case.
```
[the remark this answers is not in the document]
nicklesimba: perfect, left this in specifically for review, will do to remove it.

cmd/controlloop/controlloop.go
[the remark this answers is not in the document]
dougbtv: Also.... it's possible users passes an invalid cron syntax here, so make sure to trap that.

pkg/reconciler/errors.go
```diff
@@ -1,4 +1,4 @@
-package main
+package reconciler
```
maiqueb: not sure where to put this comment: now that you've dropped the reconciler cmd, you can also remove this file. None of these errors are used now.

pkg/controlloop/pod.go
```diff
@@ -33,11 +33,12 @@ import (
...
-	ipReconcilerQueueName = "pod-updates"
-	syncPeriod            = time.Second
-	whereaboutsConfigPath = "/etc/cni/net.d/whereabouts.d/whereabouts.conf"
-	maxRetries            = 2
+	defaultMountPath         = "/host"
+	ipReconcilerQueueName    = "pod-updates"
+	syncPeriod               = time.Second
+	whereaboutsConfigPath    = "/etc/cni/net.d/whereabouts.d/whereabouts.conf"
+	maxRetries               = 2
+	defaultReconcilerTimeout = 30 // added for reconciler invocation
```
maiqueb, line 41: isn't this the exact same thing as in https://github.com/k8snetworkplumbingwg/whereabouts/pull/238/commits/6035bb030bee3c6caed052779d7b1fa72ab1dead#diff-aa6e374b725fa31fa303df2c2516ff46f2d936054e1d8de78605a48c07d8a2dcR11 ?

Please stick to one if possible. Is there a reason for duplicating them ?

pkg/reconciler/ip.go
```diff
@@ -0,0 +1,46 @@
...
+	"time"
+
+	"github.com/k8snetworkplumbingwg/whereabouts/pkg/logging"
+)
+
+const (
+	defaultReconcilerTimeout = 30
+)
+
+func InvokeIPReconciler(returnErr chan error) {
```
maiqueb, line 14: I would use another function name - `ReconcileIPs` maybe ?

cmd/controlloop/controlloop.go
```diff
@@ -56,6 +61,26 @@ func main() {
 	}
 
 	networkController.Start(stopChan)
+
+	s := gocron.NewScheduler(time.UTC)
```
maiqueb, line 65: do we want the timezone to be configurable ? I'm OK with forcing everyone to use UTC.

pkg/reconciler/ip.go
```diff
@@ -0,0 +1,46 @@
...
+		logging.Debugf("no IP addresses to cleanup")
+	}
+
+	if err := ipReconcileLoop.ReconcileOverlappingIPAddresses(ctx); err != nil {
+		returnErr <- err
+		return
+	}
+
+	returnErr <- err
+	return
```
maiqueb: remove the redundant return statement.

pkg/reconciler/ip.go
```diff
@@ -0,0 +1,46 @@
...
+	} else {
+		logging.Debugf("no IP addresses to cleanup")
+	}
+
+	if err := ipReconcileLoop.ReconcileOverlappingIPAddresses(ctx); err != nil {
+		returnErr <- err
+		return
+	}
+
+	returnErr <- err
```
maiqueb, line 44: be explicit: `errorChannel <- nil`

maiqueb (changes requested):
Nice. Mostly I just have nits.

cmd/controlloop/controlloop.go
```diff
@@ -46,7 +49,9 @@ func main() {
 	logging.SetLogStderr(true)
 
 	stopChan := make(chan struct{})
+	returnErr := make(chan error)
```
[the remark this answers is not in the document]
nicklesimba: Sure

cmd/controlloop/controlloop.go
nicklesimba: We could change this down the line if a user finds it important. I'm indifferent but I figured UTC is standard.

pkg/controlloop/pod.go
nicklesimba: Yes it is the same exact thing. Looks like an oversight, I don't even see it referenced in this file. Good catch!

cmd/controlloop/controlloop.go
```diff
@@ -56,6 +61,26 @@ func main() {
...
+	for {
+		select {
+		case <-stopChan:
+			s.Stop()
+			return
+		case err := <-returnErr:
+			if err == nil {
+				logging.Verbosef("reconciler success")
+			} else {
+				logging.Verbosef("reconciler failure")
```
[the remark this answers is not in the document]
nicklesimba: These prints were debug prints that got left in on accident. But I like this train of thought. Also okay with removing these logging messages entirely.

cmd/controlloop/controlloop.go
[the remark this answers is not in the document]
nicklesimba: Trapping it.

pkg/controlloop/pod.go
```diff
@@ -116,9 +117,6 @@ func (pc *PodController) Start(stopChan <-chan struct{}) {
 	}
 
 	go wait.Until(pc.worker, syncPeriod, stopChan)
-
-	<-stopChan
-	logging.Verbosef("shutting down network controller")
```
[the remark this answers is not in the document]
nicklesimba: In agreement with this logic, but let's discuss how I might do this on slack.

pkg/reconciler/ip.go
nicklesimba: Sure let's go with that

pkg/controlloop/pod.go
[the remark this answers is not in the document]
nicklesimba: Done.

pkg/reconciler/controlloop/pod.go
```diff
@@ -116,9 +116,6 @@ func (pc *PodController) Start(stopChan <-chan struct{}) {
 	}
 
 	go wait.Until(pc.worker, syncPeriod, stopChan)
-
-	<-stopChan
-	logging.Verbosef("shutting down network controller")
```
maiqueb, line 121: This breaks the code-base; you need to add waiting on the stop channel to the main func.

nicklesimba: This does get addressed in the other commits - see latest code. I was trying to give this change its own commit - and in hindsight I agree adding a wait on stop channel in the main function makes this commit functional....but we don't expect or support someone to cherry pick this specific commit for their usage, so it should be fine this time.

maiqueb: Yes, but why would you choose to add a broken commit ? 

This would trip tools like [`git bisect`](https://git-scm.com/docs/git-bisect), and I don't see any advantage to have it.
 
Just behavior is even explicitly listed as a no-go in projects like [ovn-k](https://github.com/ovn-org/ovn-kubernetes/blob/master/CONTRIBUTING.md#before-you-start). 

Do you think it is needed to do the same in whereabouts ?

nicklesimba: I think I may have originally misunderstood your comments about adding a separate commit for this. I did briefly consider the repercussions of a broken commit and figured it was inconsequential. I'll see if I can revise the commit to have the "wait for stopChan" line.

nicklesimba:
Not sure if this command works upstream... /retest-required

maiqueb (changes requested):
You'll need to move the shutdown of the work-queue to the main function - it **must** run after the stopChannel is caught. 

I would simply defer it within the main. For that, you'll need to add a receiver function to the pod controller that shuts down the queue.

maiqueb (changes requested):
You now registering the fail handler, and attempting to run the specs twice in the same pkg.

You cannot do that. Remove the following lines from the `pkg/reconciler/suite_test.go` file.

```golang
func TestAPIs(t *testing.T) {
	RegisterFailHandler(Fail)
	RunSpecsWithDefaultAndCustomReporters(t,
		"Whereabouts IP reconciler Suite",
		[]Reporter{})
}
```
Or, better yet, merge the `pkg/reconciler/ip_test.go` and `pkg/reconciler/iploop_test.go` - each should have a different `Describe` block.

This would address your current unit test error.

dougbtv (approved):
LGTM! :100:

Issue #231, opened 2022-05-23
etcd deprecation


dougbtv:
Also in discussion: https://github.com/k8snetworkplumbingwg/whereabouts/discussions/228

Just looking to get any input on any community members that are still using the etcd support? Over the course of time, we've mentioned that we wouldn't be extending it, and some functionality doesn't apply to the etcd backend. 

While the extensible factor of having the multiple backends is a good idea, in practice it tends to be more work and we're not sure who's using it. Additionally, having to also manage an etcd deployment as well is often more complicated, when whereabouts tends to be more k8s focused rather than being orchestration engine agnostic, especially as it's often used in tandem with Multus CNI (which is also k8s specific).

dougbtv:
Also related to #230

maiqueb:
Plan is:
- tag a release from latest master, and create a release branch from it.
- get rid of etcd stuff on latest master.
- backport **fixes** (not features) to the etcd branch for 90 days.

Stay tuned to this issue to know when the clock starts ticking ...

maiqueb:
OK, the clock started ticking: here's the [new branch](https://github.com/k8snetworkplumbingwg/whereabouts/commits/release/v0.5-etcd).

PR #240, opened 2022-06-24
storage, etcd: remove backend

maiqueb:
Fixes: #231


dougbtv (approved):
Miguel and I discussed the method by which we're removing the etcd support, with the consideration of "what if we add another backend in the future?" -- the outcome as suggested by Miguel was that we should expose the API in such way as to not need the handler as it's used today, and I concur, it's a better approach.

dougbtv:
Additionally -- it works in my lab ;) :+1:

PR #256, opened 2022-08-04
Implement network name

### LINKED ISSUE #50: Enhancement: Allow re-use of ranges per namespace
Currently an `IPPool` will be created in the namespace as specified in the Kubernetes config (for example), so in the default install this means that the `IPPool` CRs are created in `kube-system`. 

What if we added an option to make a pool namespaced, so you could re-use a pooled range between two different namespaces?

I had a recent conversation where this was brought up. I think in the context of multi-tenancy. But, I'll return to the conversation to see if I can get input on the use-case.

toelke:

This allows specifying an optional `network_name` in the IPAM-config as such:

```json
    {                                                                                                                                                                                                                                          
      "cniVersion": "0.3.1",                                                                                                                                                                                                                   
      "name": "meganet",                                                                                                                                                                                                                       
      "type": "bridge",                                                                                                                                                                                                                        
      "bridge": "meganet",                                                                                                                                                                                                                     
      "ipam": {                                                                                                                                                                                                                                
        "type": "whereabouts",                                                                                                                                                                                                                 
        "range": "10.158.0.1/30",                                                                                                                                                                                                              
        "enable_overlapping_ranges": false,                                                                                                                                                                                                    
        "network_name": "meganet"                                                                                                                                                                                                              
      }                                                                                                                                                                                                                                        
    }                      
```

This allows re-using the same IP range for multiple L2 networks.

This is achieved by creating the `IPPool` resource in kubernetes not as `10.158.0.0-30` but as `meganet-10.158.0.0.30`. When no network-name is given, it defaults to "" and the "-" is left out to enable backwards compatibily.

Fixes #50

This is only the second time I touch go-code, so if anything seems strange, please do not hesitate to request clarification.

### DIFF
--- a/cmd/whereabouts_test.go
+++ b/cmd/whereabouts_test.go
@@ -57,7 +57,7 @@ func AllocateAndReleaseAddressesTest(ipVersion string, ipamConf *whereaboutstype
 	Expect(ipamConf.IPRanges).NotTo(BeEmpty())
 	wbClient := *kubernetes.NewKubernetesClient(
 		fake.NewSimpleClientset(
-			ipPool(ipamConf.IPRanges[0].Range, podNamespace)),
+			ipPool(ipamConf.IPRanges[0].Range, podNamespace, ipamConf.NetworkName)),
 		fakek8sclient.NewSimpleClientset(),
 		0)
 	for i := 0; i < len(expectedAddresses); i++ {
@@ -123,7 +123,7 @@ func AllocateAndReleaseAddressesTest(ipVersion string, ipamConf *whereaboutstype
 				ipamConf,
 				fakek8sclient.NewSimpleClientset(),
 				fake.NewSimpleClientset(
-					ipPool(ipamConf.IPRanges[0].Range, podNamespace)))
+					ipPool(ipamConf.IPRanges[0].Range, podNamespace, ipamConf.NetworkName)))
 			return cmdDel(args, client)
 		})
 		Expect(err).NotTo(HaveOccurred())
@@ -315,7 +315,7 @@ var _ = Describe("Whereabouts operations", func() {
 			ipamConf,
 			fakek8sclient.NewSimpleClientset(),
 			fake.NewSimpleClientset(
-				ipPool(ipamConf.IPRanges[0].Range, podNamespace)))
+				ipPool(ipamConf.IPRanges[0].Range, podNamespace, ipamConf.NetworkName)))
 
 		// Allocate the IP
 		r, raw, err := testutils.CmdAddWithArgs(args, func() error {
@@ -387,7 +387,7 @@ var _ = Describe("Whereabouts operations", func() {
 			ipamConf,
 			fakek8sclient.NewSimpleClientset(),
 			fake.NewSimpleClientset(
-				ipPool(ipamConf.IPRanges[0].Range, podNamespace)))
+				ipPool(ipamConf.IPRanges[0].Range, podNamespace, ipamConf.NetworkName)))
 
 		// Allocate the IP
 		r, raw, err := testutils.CmdAddWithArgs(args, func() error {
@@ -456,7 +456,7 @@ var _ = Describe("Whereabouts operations", func() {
 			ipamConf,
 			fakek8sclient.NewSimpleClientset(),
 			fake.NewSimpleClientset(
-				ipPool(ipamConf.IPRanges[0].Range, podNamespace)))
+				ipPool(ipamConf.IPRanges[0].Range, podNamespace, ipamConf.NetworkName)))
 
 		// Allocate the IP
 		r, raw, err := testutils.CmdAddWithArgs(args, func() error {
@@ -536,7 +536,7 @@ var _ = Describe("Whereabouts operations", func() {
 			ipamConf,
 			fakek8sclient.NewSimpleClientset(),
 			fake.NewSimpleClientset(
-				ipPool(ipamConf.IPRanges[0].Range, podNamespace)))
+				ipPool(ipamConf.IPRanges[0].Range, podNamespace, ipamConf.NetworkName)))
 
 		// Allocate the IP
 		r, raw, err := testutils.CmdAddWithArgs(args, func() error {
@@ -625,7 +625,7 @@ var _ = Describe("Whereabouts operations", func() {
 			ipamConf,
 			fakek8sclient.NewSimpleClientset(),
 			fake.NewSimpleClientset(
-				ipPool(ipamConf.IPRanges[0].Range, podNamespace)))
+				ipPool(ipamConf.IPRanges[0].Range, podNamespace, ipamConf.NetworkName)))
 
 		// Allocate the IP
 		r, raw, err := testutils.CmdAddWithArgs(args, func() error {
@@ -689,8 +689,8 @@ var _ = Describe("Whereabouts operations", func() {
 			ipamConf,
 			fakek8sclient.NewSimpleClientset(),
 			fake.NewSimpleClientset(
-				ipPool(ipamConf.IPRanges[0].Range, podNamespace),
-				ipPool(ipamConf.IPRanges[1].Range, podNamespace)))
+				ipPool(ipamConf.IPRanges[0].Range, podNamespace, ipamConf.NetworkName),
+				ipPool(ipamConf.IPRanges[1].Range, podNamespace, ipamConf.NetworkName)))
 
 		// Allocate the IP
 		r, raw, err := testutils.CmdAddWithArgs(args, func() error {
@@ -754,8 +754,8 @@ var _ = Describe("Whereabouts operations", func() {
 			ipamConf,
 			fakek8sclient.NewSimpleClientset(),
 			fake.NewSimpleClientset(
-				ipPool(ipamConf.IPRanges[0].Range, podNamespace),
-				ipPool(ipamConf.IPRanges[1].Range, podNamespace)))
+				ipPool(ipamConf.IPRanges[0].Range, podNamespace, ipamConf.NetworkName),
+				ipPool(ipamConf.IPRanges[1].Range, podNamespace, ipamConf.NetworkName)))
 
 		// Allocate the IP
 		r, raw, err := testutils.CmdAddWithArgs(args, func() error {
@@ -817,7 +817,7 @@ var _ = Describe("Whereabouts operations", func() {
 			ipamConf,
 			fakek8sclient.NewSimpleClientset(),
 			fake.NewSimpleClientset(
-				ipPool(ipamConf.IPRanges[0].Range, podNamespace)))
+				ipPool(ipamConf.IPRanges[0].Range, podNamespace, ipamConf.NetworkName)))
 
 		// Allocate the IP
 		r, raw, err := testutils.CmdAddWithArgs(args, func() error {
@@ -883,7 +883,7 @@ var _ = Describe("Whereabouts operations", func() {
 			ipamConf,
 			fakek8sclient.NewSimpleClientset(),
 			fake.NewSimpleClientset(
-				ipPool(ipamConf.IPRanges[0].Range, podNamespace)))
+				ipPool(ipamConf.IPRanges[0].Range, podNamespace, ipamConf.NetworkName)))
 
 		// Allocate the IP
 		r, raw, err := testutils.CmdAddWithArgs(args, func() error {
@@ -939,7 +939,7 @@ var _ = Describe("Whereabouts operations", func() {
 		Expect(ipamConf.IPRanges).NotTo(BeEmpty())
 		wbClient := *kubernetes.NewKubernetesClient(
 			fake.NewSimpleClientset(
-				ipPool(ipamConf.IPRanges[0].Range, podNamespace)),
+				ipPool(ipamConf.IPRanges[0].Range, podNamespace, ipamConf.NetworkName)),
 			fakek8sclient.NewSimpleClientset(),
 			0)
 
@@ -1036,7 +1036,7 @@ var _ = Describe("Whereabouts operations", func() {
 		Expect(ipamConf.IPRanges).NotTo(BeEmpty())
 		wbClient := *kubernetes.NewKubernetesClient(
 			fake.NewSimpleClientset(
-				ipPool(ipamConf.IPRanges[0].Range, podNamespace)),
+				ipPool(ipamConf.IPRanges[0].Range, podNamespace, ipamConf.NetworkName)),
 			fakek8sclient.NewSimpleClientset(),
 			0)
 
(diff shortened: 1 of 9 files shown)


maiqueb:
I don't think this fixes #50 - that issues clearly mentions `namespaced pools`. This allows the same range to be spawned across multiple L2 networks, something for which I don't see a clear use case.

Could you elaborate on what use case is this enabling ?

toelke:
You may be right there -- but the last comment on #50 mentions this idea: https://github.com/k8snetworkplumbingwg/whereabouts/issues/50#issuecomment-874040513

In case you are running multiple L2 networks (by using VLANs, VxLANs, ...) you might want to allow assigning IP addresses to those networks independently. This PR allows to do that.

toelke:
I marked this as draft until the conclusion of #258 which might require changes to this PR

toelke:
@dougbtv @maiqueb I am not sure about the failing e2e tests -- is that related to this PR?

toelke:
I have rebased this branch and will look at the test results.

The discussion in https://github.com/k8snetworkplumbingwg/whereabouts/pull/258#discussion_r1043503332 also leads me to question the choice of name for the new field:

* the implementation as it is now calls it `network_name`
* Doug called it `pool_name` in his example
* I also like `address_collision_domain` as it is extremely explicit

We should also think about Ayushs idea of defining sematics for the comparison of `IPPool`s -- although I think that that might become a secondary feature to refine the ease of use of this "Overlapping Ranges" feature.

toelke:
I have "fixed" the e2e tests in a way that I consider to be relatively hacky. I clearly am to inexperienced in go to understand the structure of the tests. I have not found the place where the `network-attachment-definitions` are created and where I could add the name.

maiqueb:
> I have "fixed" the e2e tests in a way that I consider to be relatively hacky. I clearly am to inexperienced in go to understand the structure of the tests. I have not found the place where the `network-attachment-definitions` are created and where I could add the name.

That's OK; I'll guide you with that. Let me check your code, and I'll point you to the correct place. 

Thanks for the work so far.

toelke:
Also, I guess we should add at least one e2e test for actually running two `IPPools` with different names

pkg/storage/kubernetes/ipam.go
```diff
@@ -116,7 +116,7 @@ func (i *KubernetesIPAM) GetIPPool(ctx context.Context, ipRange string) (storage
 	return &KubernetesIPPool{i.client, i.containerID, firstIP, pool}, nil
 }
 
-func NormalizeRange(ipRange string) string {
+func NormalizeRange(ipRange string, networkName string) string {
```
maiqueb, line 119: hm, we need to come up with a better way to do this: this function normalized a range before, but actually, what this did was generate the name of the `IPPool`. That's made more obvious now.

I propose we make this clearer (would appreciate a better name for the `GlobalIPPoolName` though):
```golang
func GlobalIPPoolName(ipRange string) string {
    return normalizedRange(ipRange)
}

func IPPoolName(networkName string, ipRange string) string {
    return fmt.Sprintf("%s-%s", networkName, normalizedRange(ipRange))
}

func normalizedRange(ipRange string) string {
	if ipRange[len(ipRange)-1] == ':' {
		ipRange = ipRange + "0"
	}
	normalized := strings.ReplaceAll(ipRange, ":", "-")
	normalized = strings.ReplaceAll(normalized, "/", "-")
	return normalized
}
```

And then let the caller choose which to use - the production code caller can stay as you have (well, it'll need to check which of the 2 funcs to call ...), but the e2e tests tests will IMO be a lot easier to follow.

cmd/whereabouts_test.go
```diff
@@ -57,7 +57,7 @@ func AllocateAndReleaseAddressesTest(ipVersion string, ipamConf *whereaboutstype
 	Expect(ipamConf.IPRanges).NotTo(BeEmpty())
 	wbClient := *kubernetes.NewKubernetesClient(
 		fake.NewSimpleClientset(
-			ipPool(ipamConf.IPRanges[0].Range, podNamespace)),
+			ipPool(ipamConf.IPRanges[0].Range, podNamespace, ipamConf.NetworkName)),
```
maiqueb, line 60: I'm confused; shouldn't this the the "global" pool ?

e2e/client/ippool.go
```diff
@@ -13,9 +13,9 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-func isIPPoolAllocationsEmpty(k8sIPAM *kubeClient.KubernetesIPAM, ipPoolName string) wait.ConditionFunc {
+func isIPPoolAllocationsEmpty(k8sIPAM *kubeClient.KubernetesIPAM, ipPoolCIDR string) wait.ConditionFunc {
```
maiqueb, line 16: You'll need to parameterize this function with the network name, since I expect you to add a new e2e test where the same range can be used by different networks.

I think you should add a new type (struct) featuring both the CIDR and the name. Then pass this around.

I think its idiomatic, and fits well with the chosen model.

e2e/client/ippool.go
```diff
@@ -33,6 +33,6 @@ func isIPPoolAllocationsEmpty(k8sIPAM *kubeClient.KubernetesIPAM, ipPoolName str
 
 // WaitForZeroIPPoolAllocations polls up to timeout seconds for IP pool allocations to be gone from the Kubernetes cluster.
 // Returns an error if any IP pool allocations remain after time limit, or if GETing IP pools causes an error.
-func WaitForZeroIPPoolAllocations(k8sIPAM *kubeClient.KubernetesIPAM, ipPoolName string, timeout time.Duration) error {
-	return wait.PollImmediate(time.Second, timeout, isIPPoolAllocationsEmpty(k8sIPAM, ipPoolName))
+func WaitForZeroIPPoolAllocations(k8sIPAM *kubeClient.KubernetesIPAM, ipPoolCIDR string, timeout time.Duration) error {
```
maiqueb, line 36: ditto

e2e/e2e_test.go
```diff
@@ -282,7 +282,7 @@ var _ = Describe("Whereabouts functionality", func() {
 					Expect(err).NotTo(HaveOccurred())
 					Expect(podList.Items).NotTo(BeEmpty())
 
-					ipPool, err := k8sIPAM.GetIPPool(context.Background(), ipPoolName)
+					ipPool, err := k8sIPAM.GetIPPool(context.Background(), ipPoolCIDR, "")
```
maiqueb, line 285: please use a constant instead of the magic string `""`. Name it something like `uniquePerCIDRPoolName` or something like that. 

I would welcome any other name ... That one was not good :)

pkg/storage/kubernetes/ipam.go
```diff
@@ -100,8 +100,8 @@ func toAllocationMap(reservelist []whereaboutstypes.IPReservation, firstip net.I
 }
 
 // GetIPPool returns a storage.IPPool for the given range
-func (i *KubernetesIPAM) GetIPPool(ctx context.Context, ipRange string) (storage.IPPool, error) {
-	normalized := NormalizeRange(ipRange)
+func (i *KubernetesIPAM) GetIPPool(ctx context.Context, ipRange string, networkName string) (storage.IPPool, error) {
```
maiqueb, line 103: Please make this func receive as input a new type (struct) featuring both the CIDR and the name.

This should be reused in the e2e tests / unit tests.

pkg/controlloop/pod.go
```diff
@@ -271,8 +271,8 @@ func (pc *PodController) ifaceNetAttachDef(ifaceStatus nadv1.NetworkStatus) (*na
 	return nad, nil
 }
 
-func (pc *PodController) ipPool(cidr string) (*whereaboutsv1alpha1.IPPool, error) {
-	pool, err := pc.ipPoolLister.IPPools(ipPoolsNamespace()).Get(wbclient.NormalizeRange(cidr))
+func (pc *PodController) ipPool(cidr string, networkName string) (*whereaboutsv1alpha1.IPPool, error) {
```
maiqueb, line 274: likewise, use this new struct here.

pkg/controlloop/entity_generators.go
```diff
@@ -96,10 +96,10 @@ func podNetworkStatusAnnotations(namespace string, networkNames ...string) strin
 	return string(serelizedNetStatus)
 }
 
-func ipPool(ipRange string, namespace string, podReferences ...string) *v1alpha1.IPPool {
+func ipPool(ipRange string, networkName string, namespace string, podReferences ...string) *v1alpha1.IPPool {
```
maiqueb, line 99: use this new struct as parameter here

maiqueb (changes requested):
Nice !

You'll need to add a new e2e test with 2 networks having the same CIDR (which would lead to 2 different pools). Then, create a pod on each of the networks, and ensure both pods got the same IP address.

cmd/whereabouts_test.go
toelke: This passes through whatever is configured in the test in the `ipamConf` -- if a test uses `""` it uses the "global" (unnamed?) pool. But this is consistent as currently "normalizing" with an empty Name will give the unnamed pool.

pkg/storage/kubernetes/ipam.go
toelke: ✅

e2e/client/ippool.go
toelke: ✅

e2e/e2e_test.go
toelke: ✅

pkg/storage/kubernetes/ipam.go
toelke: ✅

pkg/controlloop/pod.go
toelke: ✅

pkg/controlloop/entity_generators.go
toelke: ✅

toelke: ✅

pkg/controlloop/pod.go
toelke: ✅

toelke:
> You'll need to add a new e2e test with 2 networks having the same CIDR (which would lead to 2 different pools). Then, create a pod on each of the networks, and ensure both pods got the same IP address.

✅

maiqueb (changes requested):
neat !!! 

Really close. 

I would just use as a parameter this new struct you've added on all the e2e test helpers.

And I'm a bit confused on the docs as well.

Thank you!

e2e/client/whereabouts.go
```diff
@@ -68,7 +68,7 @@ func (c *ClientInfo) ProvisionPod(podName string, namespace string, label, annot
 		return nil, err
 	}
 
-	const podCreateTimeout = 10 * time.Second
+	const podCreateTimeout = 60 * time.Second
```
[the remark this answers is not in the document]
toelke: Because on my test-machine some tests took longer than that. I absolutely did not want to commit that. :man_facepalming:

doc/proposals/named-networks.md
```diff
@@ -1,5 +1,7 @@
 # Support for named networks
 
+It was decided in March 2023 to implement scheme 3. This means that we added a new field to the `IPAMConfig` called `network_name` that is included in the `IPPool` name to distinguish it from other pools with the same CIDR.
```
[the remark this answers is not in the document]
toelke: Yes, that was what I proposed, it was not what was decided.

e2e/client/ippool.go
```diff
@@ -13,9 +13,9 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-func isIPPoolAllocationsEmpty(k8sIPAM *kubeClient.KubernetesIPAM, ipPoolName string) wait.ConditionFunc {
+func isIPPoolAllocationsEmpty(k8sIPAM *kubeClient.KubernetesIPAM, ipPoolCIDR string) wait.ConditionFunc {
 	return func() (bool, error) {
-		ipPool, err := k8sIPAM.GetIPPool(context.Background(), ipPoolName)
+		ipPool, err := k8sIPAM.GetIPPool(context.Background(), kubeClient.PoolIdentifier{IpRange: ipPoolCIDR, NetworkName: kubeClient.UnnamedNetwork})
```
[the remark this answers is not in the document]
toelke: For no current test we need to check a named pool.

e2e/e2e_test.go
```diff
@@ -477,6 +481,73 @@ var _ = Describe("Whereabouts functionality", func() {
...
+				netAttachDef2 *nettypes.NetworkAttachmentDefinition
+				pod2          *core.Pod
+			)
+
+			BeforeEach(func() {
+				var (
+					err error
+				)
+
+				netAttachDef2 = macvlanNetworkWithWhereaboutsIPAMNetwork(testNetworkName+"2", testNamespace, ipv4TestRange, []string{}, testNetworkName+"2", false)
```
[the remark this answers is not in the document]
toelke: ✅

e2e/e2e_test.go
```diff
@@ -493,7 +494,7 @@ var _ = Describe("Whereabouts functionality", func() {
 					err error
 				)
 
-				netAttachDef2 = macvlanNetworkWithWhereaboutsIPAMNetwork(testNetworkName+"2", testNamespace, ipv4TestRange, []string{}, testNetworkName+"2", false)
+				netAttachDef2 = macvlanNetworkWithWhereaboutsIPAMNetwork(testNetwork2Name, testNamespace, ipv4TestRange, []string{}, testNetwork2Name, false)
```
maiqueb, line 497: please define this const in this `Context` block. It doesn't need to be available to no one else.

maiqueb (changes requested):
OK I think this is good enough after you define the const in the proper block.

Would you mind squashing all commits into a single one ? And come up w/ a good commit message.

cmd/whereabouts_test.go
maiqueb: OK, I guess this is explicit enough.

e2e/e2e_test.go
```diff
@@ -506,6 +507,7 @@ var _ = Describe("Whereabouts functionality", func() {
 
 			BeforeEach(func() {
 				const singlePodName = "whereabouts-basic-test"
+				const singlePod2Name = "whereabouts-basic-test-2"
```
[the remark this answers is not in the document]
toelke: ✅

maiqueb (approved):
Thank you for your patience.

Giving @dougbtv a chance to review this as well.

dougbtv (approved):
Thank you for your awesome collaboration with the community and for going through the design process with us. Much appreciated and appreciate the hard work (and Miguel's help on the reviews!)

toelke:
Thank you so much for your guidance through the process!

PR #257, opened 2022-08-04
fix overwriting of OverlappingRanges by mergo

toelke:

When configuring `enable_overlapping_ranges` in the pod-configuration (e.g. by using a multus `NetworkAttachmentDefinition`, the parsing of the configuration makes an error. This was already noticed in #164 but I did not manage to get that solution to work.

What happens is that parsing the flatfile (where `enable_overlapping_ranges` is not set) results in a in-memory representation of the configuration where `OverlappingRanges` is set to `true`; this will overwrite the pod-specific configuration of `false`.


maiqueb (changes requested):
Looks good.

We'd need a unit test to verify it does what we want it to though.

toelke:
I have expanded an existing test. It indeed fails before the fix and is successful afterwards.

maiqueb:
> I have expanded an existing test. It indeed fails before the fix and is successful afterwards.

I would still prefer to see a new unit test, focused explicitly on this - that is easier to follow, and will work as documentation of what we expect around this flag. 

Modifying an existing test does the job, but it does not work for well for documenting the behavior. 
Since we're doing BDD on (most of) the tests, I think that should be the focus.

I would add something like:
```golang
        var globalConf string

        BeforeEach(func() {
        		globalConf = `{
      "datastore": "kubernetes",
      "kubernetes": {
        "kubeconfig": "/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig"
      },
      "log_file": "/tmp/whereabouts.log",
      "log_level": "debug",
      "gateway": "192.168.5.5"
    }`

		Expect(ioutil.WriteFile("/tmp/whereabouts.conf", []byte(globalconf), 0755)).To(Succeed())
        })
        
	It("overlapping range can be set", func() {
		ipamconfig, _, err := LoadIPAMConfig([]byte(generateIPAMConfWithOverlappingRanges()), "")
		
		// Expect the flatfile config not to have overridden the config in the NAD
		Expect(ipamconfig.OverlappingRanges).To(BeTrue())
	})
	
	It("overlapping range can be disabled", func() {
		ipamconfig, _, err := LoadIPAMConfig([]byte(generateIPAMConfWithoutOverlappingRanges()), "")
		
		// Expect the flatfile config not to have overridden the config in the NAD
		Expect(ipamconfig.OverlappingRanges).To(BeFalse())
	})

func generateIPAMConfWithOverlappingRanges() string {
	return `{
      "cniVersion": "0.3.1",
      "name": "mynet",
      "type": "ipvlan",
      "master": "foo0",
      "ipam": {
        "configuration_path": "/tmp/whereabouts.conf",
        "type": "whereabouts",
        "range": "192.168.2.230/24",
        "range_start": "192.168.2.223",
        "gateway": "192.168.10.1",
        "leader_lease_duration": 3000,
        "leader_renew_deadline": 2000,
        "leader_retry_period": 1000
        "leader_retry_period": 1000,
        "enable_overlapping_ranges": true
      }
      }`
}

func generateIPAMConfWithoutOverlappingRanges() string {
	return `{
      "cniVersion": "0.3.1",
      "name": "mynet",
      "type": "ipvlan",
      "master": "foo0",
      "ipam": {
        "configuration_path": "/tmp/whereabouts.conf",
        "type": "whereabouts",
        "range": "192.168.2.230/24",
        "range_start": "192.168.2.223",
        "gateway": "192.168.10.1",
        "leader_lease_duration": 3000,
        "leader_renew_deadline": 2000,
        "leader_retry_period": 1000
        "leader_retry_period": 1000,
        "enable_overlapping_ranges": false
      }
      }`
}
```

You could even use a table for these 2 tests, thus avoiding the duplicate test code (pass the configuration and the matcher as inputs to the func, for instance).

toelke:
You are correct. I also changed your proposed test to actually overwrite the flatfile in both directions -- the original code does not allow overwriting a `true` with a `false`. The fixed code allows both directions.

pkg/config/config_test.go
```diff
@@ -94,7 +124,44 @@ var _ = Describe("Allocation operations", func() {
...
+			"datastore": "kubernetes",
+			"kubernetes": {
+				"kubeconfig": "/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig"
+			},
+			"log_file": "/tmp/whereabouts.log",
+			"log_level": "debug",
+			"gateway": "192.168.5.5",
+			"enable_overlapping_ranges": false
+		}`
+		Expect(ioutil.WriteFile("/tmp/whereabouts.conf", []byte(globalConf), 0755)).To(Succeed())
```
maiqueb, line 140: Seems to me we're leaking this file after the execution, but I've noticed that also happens for other tests.

I'll handle that in a follow-up.

maiqueb (changes requested):

maiqueb:
You should rebase latest master, to give the e2e tests a chance to work - with the current state of your tree, it'll fail trying to install the CNI plugins in the k8s nodes.

toelke:
Done

maiqueb (approved):
Had forgotten to hit the approve button - sorry for that :)

PR #258, opened 2022-08-04
Proposal for named networks

toelke:
As requested [here](https://github.com/k8snetworkplumbingwg/whereabouts/issues/50#issuecomment-1205035504) this is a proposal for named networks to allow using the same CIDR range for multiple networks (e.g. in multi-tenancy situations).

Cc @maiqueb

### DIFF
--- a/doc/proposals/named-networks.md
+++ b/doc/proposals/named-networks.md
@@ -211,7 +211,7 @@ We propose to supply a small program (ideally also as OCI image for use in kuber
 
 This tool would do roughly the following:
 
-0. Caution the user that no new Pods with whereabouts CNI-config should be started for the duration of the run
+0. Grab the leader election lock to prevent any whereabouts to run during the operation
 1. List all `IPPool`s in the `kube-system` namespace
 2. List all `NetworkAttachmentDefinition`s in all namespaces
 3. Match the `NetworkAttachmentDefinition`s to the `IPPool`s using the CIDR canonicalization rules
--- a/doc/proposals/named-networks.md
+++ b/doc/proposals/named-networks.md
@@ -1,0 +1,185 @@
+# Support for named networks
+
+# Table of contents
+
+- [Introduction](#introduction)
+  - [Goal](#goal-of-this-proposal)
+- [Design](#design)
+  - [Changes in IPAM Config](#changes-in-ipam-config)
+  - [Changes in Modules](#changes-in-modules)
+- [Summary](#summary)
+- [Discussions and Decisions](#discussions-and-decisions)
+
+<hr>
+
+## Introduction
+
+When whereabouts assigns an IP to a Pod this fact is recorded in a document of kind `IPPool` that has its name derived from the CIDR range in question.
+Should the user configure multiple overlapping ranges, it is possible to configure whereabouts to allow assigning duplicate IPs.
+However, since the storage of the assignments is done in a document that is named like the CIDR range, it is not possible to configure *the same CIDR range* twice and have whereabouts assign from the ranges independently.
+
+This is, for example, useful in multi-tenant situations where more than one group is responsible for selecting CIDR ranges.
+
+### Goal of this Proposal
+
+- Allow configuring the same CIDR range multiple times (e.g. in separate multus-`NetworkAttachmentDefinitions`)
+
+<hr>
+
+## Design
+
+The IPAM configuration format would be updated to include an optional field for the name of the network in question.
+If this field is left empty, whereabouts will behave as it does now, without these changes.
+
+### Changes in IPAM Config
+
+This will lead to change in IPAM config something as follows:
+
+<table>
+<tr>
+<th>Old IPAM Config</th>
+<th>Changes</th>
+<th>New IPAM Config</th>
+</tr>
+<tr>
+<td>
+  
+```json
+{
+      "cniVersion": "0.3.0",
+      "name": "whereaboutsexample",
+      "type": "macvlan",
+      "master": "eth0",
+      "mode": "bridge",
+      "ipam": {
+        "type": "whereabouts",
+        "range": "192.168.2.225/28",
+        "exclude": [
+           "192.168.2.229/30",
+           "192.168.2.236/32"
+        ]
+      }
+}
+```
+  
+</td>
+<td>
+
+```diff
+{
+      "cniVersion": "0.3.0",
+      "name": "whereaboutsexample",
+      "type": "macvlan",
+      "master": "eth0",
+      "mode": "bridge",
+      "ipam": {
+        "type": "whereabouts",
+        "range": "192.168.2.225/28",
++       "network_name": "whereaboutsexample",
+        "exclude": [
+           "192.168.2.229/30",
+           "192.168.2.236/32"
+        ]
+      }
+}
+```
+
+</td>
+<td>
+
+```json
+{
+      "cniVersion": "0.3.0",
+      "name": "whereaboutsexample",
+      "type": "macvlan",
+      "master": "eth0",
+      "mode": "bridge",
+      "ipam": {
+        "type": "whereabouts",
+        "range": "192.168.2.225/28",
+        "network_name": "whereaboutsexample",
+        "exclude": [
+           "192.168.2.229/30",
+           "192.168.2.236/32"
+        ]
+      }
+}
+```
+
+</td>
+</tr>
+</table>
+
+### Changes in Modules
+
+#### whereabouts/pkg/types/types.go
+
+```diff
+
+ type IPAMConfig struct {
+      Name                string
+      Type                string               `json:"type"`
+      Routes              []*cnitypes.Route    `json:"routes"`
+      Datastore           string               `json:"datastore"`
+      Addresses           []Address            `json:"addresses,omitempty"`
+      OmitRanges          []string             `json:"exclude,omitempty"`
+      DNS                 cnitypes.DNS         `json:"dns"`
+      Range               string               `json:"range"`
+      RangeStart          net.IP               `json:"range_start,omitempty"`
+      RangeEnd            net.IP               `json:"range_end,omitempty"`
+      GatewayStr          string               `json:"gateway"`
+      EtcdHost            string               `json:"etcd_host,omitempty"`
+      EtcdUsername        string               `json:"etcd_username,omitempty"`
+      EtcdPassword        string               `json:"etcd_password,omitempty"`
+      EtcdKeyFile         string               `json:"etcd_key_file,omitempty"`
+      EtcdCertFile        string               `json:"etcd_cert_file,omitempty"`
+      EtcdCACertFile      string               `json:"etcd_ca_cert_file,omitempty"`
+      LeaderLeaseDuration int                  `json:"leader_lease_duration,omitempty"`
+      LeaderRenewDeadline int                  `json:"leader_renew_deadline,omitempty"`
+      LeaderRetryPeriod   int                  `json:"leader_retry_period,omitempty"`
+      LogFile             string               `json:"log_file"`
+      LogLevel            string               `json:"log_level"`
+      OverlappingRanges   bool                 `json:"enable_overlapping_ranges,omitempty"`
+      SleepForRace        int                  `json:"sleep_for_race,omitempty"`
+      Gateway             net.IP
+      Kubernetes          KubernetesConfig     `json:"kubernetes,omitempty"`
+      ConfigurationPath   string               `json:"configuration_path"`
+      PodName             string
+      PodNamespace        string
+      NetworkName         string               `json:"network_name,omitempty"`
+ }
+```
+
+#### whereabouts/pkg/storage/kubernetes/ipam.go
+
+```diff
+-func NormalizeRange(ipRange string) string {
++func NormalizeRange(ipRange string, networkName string) string {
+    // v6 filter
+    normalized := strings.ReplaceAll(ipRange, ":", "-")
+    // replace subnet cidr slash
+    normalized = strings.ReplaceAll(normalized, "/", "-")
+-   return normalized
+
++   if networkName != "" {
++       return networkName + "-" + normalized
++   } else {
++       return normalized
++   }
+}
+```
+
+This will ensure that every time whereabouts looks up the current assignments on a range, it queries not for `192.168.2.225-28` but for `whereaboutsexample-192.168.2.225-28`.
+Should the network name be left empty, the lookup is for the unchanged name `192.168.2.225-28`.
+
+### Summary
+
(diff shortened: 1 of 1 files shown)


maiqueb:
Thank you. 

I'll get to this next week.

toelke:
If there is anything I can do to support you please let me know!

doc/proposals/named-networks.md
```diff
@@ -0,0 +1,185 @@
...
+### Goal of this Proposal
+
+- Allow configuring the same CIDR range multiple times (e.g. in separate multus-`NetworkAttachmentDefinitions`)
+
+<hr>
+
+## Design
+
+The IPAM configuration format would be updated to include an optional field for the name of the network in question.
+If this field is left empty, whereabouts will behave as it does now, without these changes.
```
maiqueb, line 32: Why do we need to add a new field ? 

AFAIU, the `IPAMConf` struct already features the name attribute, which just [carries over](https://github.com/k8snetworkplumbingwg/whereabouts/blob/545f05525c710a1f7f5b34c47ee6396d223581bc/pkg/config/config.go#L158) the name of the whereabouts `net` struct.

doc/proposals/named-networks.md
```diff
@@ -0,0 +1,185 @@
...
+}
+```
+  
+</td>
+<td>
+
+```diff
+{
+      "cniVersion": "0.3.0",
+      "name": "whereaboutsexample",
```
maiqueb, line 71: Why don't you use this one ?... Why would we need to add yet another attribute to the `IPAMConf` struct when this one is already available ?

doc/proposals/named-networks.md
```diff
@@ -0,0 +1,185 @@
...
+      "type": "macvlan",
+      "master": "eth0",
+      "mode": "bridge",
+      "ipam": {
+        "type": "whereabouts",
+        "range": "192.168.2.225/28",
++       "network_name": "whereaboutsexample",
```
maiqueb, line 78: I don't think this is needed. What's wrong with using the `name` attribute defined [above](https://github.com/k8snetworkplumbingwg/whereabouts/blob/545f05525c710a1f7f5b34c47ee6396d223581bc/pkg/types/types.go#L44) ?

doc/proposals/named-networks.md
```diff
@@ -0,0 +1,185 @@
...
+      LogFile             string               `json:"log_file"`
+      LogLevel            string               `json:"log_level"`
+      OverlappingRanges   bool                 `json:"enable_overlapping_ranges,omitempty"`
+      SleepForRace        int                  `json:"sleep_for_race,omitempty"`
+      Gateway             net.IP
+      Kubernetes          KubernetesConfig     `json:"kubernetes,omitempty"`
+      ConfigurationPath   string               `json:"configuration_path"`
+      PodName             string
+      PodNamespace        string
+      NetworkName         string               `json:"network_name,omitempty"`
```
maiqueb, line 149:
```suggestion
```

maiqueb (changes requested):
I would like to see you study an alternative in which we break the CIDR as the `IPPool` name.

To me, that's the most confusing / weirdest part of the API.

The idea would be to use a regular name as name (... because it's what I think a regular person would expect) and just use the `IPPool.Spec.Range` attribute to hold the range.

Querying the pool would happen also by name, so I wouldn't expect many changes there.

... not exactly sure how would debugging a wrong configuration go, though... 

While saying this, I do see value in the multi-tenancy aspect. I just think this makes even more confusing the weirdest part of our API.

toelke:
I have added a section to this proposal analyzing different ways we could generate the new name. I think the open change requests are all addressed by that. Could you please Resolve them if that is the case?

Which of the three proposals would you like to research further?

maiqueb (changes requested):

doc/proposals/named-networks.md
```diff
@@ -0,0 +1,187 @@
...
+</tr>
+<tr>
+<td>
+:green_circle: Clean design<br/>
+:green_circle: Ranges are easy to find during debugging<br/>
+:green_circle: No more IP-to-string canonicalization<br/>
+</td>
+<td>
+:red_circle: Not backwards compatible, existing installation would need to carefully migrate the existing `IPPool`s to not get duplicate IPs<br/>
+:red_circle: Unclear semantics when two ranges with the same name but different CIDR-ranges are created<br/>
```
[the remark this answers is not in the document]
toelke: > how can you have two ranges with the same name in the same namespace ?

In my testing all `IPPools` where always created in the `kube-system`-namespace -- am I misremembering things?

> I am assuming a new pod will only have access to net-attach-defs in its namespace. I would personally even favor a solution where it wouldn't be able to access the cluster's default namespace - unless the pod was created there.

For our use-case we actually need to be able to (also) have central shared `NetworkAttachmentConfiguration`s.

I will re-write the proposal:
* `IPPools` are created in the namespace of the `NetworkAttachmentDefinition`
* Cross-Namespace access is allowed if multus allows it

doc/proposals/named-networks.md
```diff
@@ -0,0 +1,187 @@
...
+<td>
+:green_circle: Backwards compatible, existing `IPPool`s are still where we left them<br/>
+:yellow_circle: Named ranges are easy to find during debugging, other ranges are unchanged<br/>
+</td>
+<td>
+:red_circle: "API" change<br/>
+</td>
+</tr>
+</table>
+
```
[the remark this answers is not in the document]
toelke: Ah; I thought that choice was to be one of the "community"?

Personally, I would have chosen the option that I already implemented -- I am usually more concerned about easy-to-understand changes than architectural cleanliness...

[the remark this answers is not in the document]
maiqueb: The choice is, but you should point out - and explain why - what is the best choice.

doc/proposals/named-networks.md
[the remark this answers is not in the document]
maiqueb: > > how can you have two ranges with the same name in the same namespace ?
> 
> In my testing all `IPPools` where always created in the `kube-system`-namespace -- am I misremembering things?

That's correct, but I don't see how this invalidates my claim: you list as a drawback "Unclear semantics when two ranges with the same name but different CIDR-ranges are created"

Hence the question: how can you have a pool with the same in the same common namespace ? Maybe I'm missing something, but I don't understand your point. I'm assuming this drawback applies to the option where we redesign the pools to have a name instead of a range.

> 
> > I am assuming a new pod will only have access to net-attach-defs in its namespace. I would personally even favor a solution where it wouldn't be able to access the cluster's default namespace - unless the pod was created there.
> 
> For our use-case we actually need to be able to (also) have central shared `NetworkAttachmentConfiguration`s.
> 
> I will re-write the proposal:
> 
> * `IPPools` are created in the namespace of the `NetworkAttachmentDefinition`
> * Cross-Namespace access is allowed if multus allows it

Make sure to add this to a requirement section / something alike.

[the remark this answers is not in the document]
toelke: >> In my testing all IPPools where always created in the kube-system-namespace -- am I misremembering things?
>
> That's correct, but I don't see how this invalidates my claim: you list as a drawback "Unclear semantics when two ranges with the same name but different CIDR-ranges are created"

You can create two `NetworkAttachmentDefinition`s called `example` in two namespaces -- whereabouts would then try to create two `IPPools` called `example` -- possibly with different CIDRs defined.

So if we go with the "only the name of the network goes into the IPPool-name", then we should possibly put the IPPool in the same namespace as the `NetworkAttachmentDefinition`?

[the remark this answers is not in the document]
maiqueb: > So if we go with the "only the name of the network goes into the IPPool-name", then we should possibly put the IPPool in the same namespace as the `NetworkAttachmentDefinition`?

I had assumed that, yes.

Is there any particular reason to keep all the pools in the same namespace ? @dougbtv

toelke:
I have updated the document with the pending changes.

doc/proposals/named-networks.md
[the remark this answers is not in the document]
toelke: :heavy_check_mark:

dougbtv:
@toelke -- thanks for the proposal! Would you be willing to join one of our community meetings to discuss? I'm really interested in the feature, and we'd love to have your efforts when you're back in october.

Meeting times and agendas @ https://github.com/k8snetworkplumbingwg/community#meetings if you can find any that work for you.

Probably the Multus maintainer's meeting would be the ideal place to talk details. But also the main Network Plumbing Working Group meeting would also work.

Thanks!

kmlebedev:
> @toelke -- thanks for the proposal! Would you be willing to join one of our community meetings to discuss? I'm really interested in the feature, and we'd love to have your efforts when you're back in october.
> 
> Meeting times and agendas @ https://github.com/k8snetworkplumbingwg/community#meetings if you can find any that work for you.
> 
> Probably the Multus maintainer's meeting would be the ideal place to talk details. But also the main Network Plumbing Working Group meeting would also work.
> 
> Thanks!

@dougbtv Maybe it's possible to have a meeting at 4:30 PM - this will be optimal for the main continents
image

toelke:
I have added an entry to the Agenda for Thursday 20th (day after tomorrow).

toelke:
It seems the meeting on 2022-11-03 did not happen; I will try attending the next one.

maiqueb:
> It seems the meeting on 2022-11-03 did not happen; I will try attending the next one.

Oh but it did; check the minutes in https://docs.google.com/document/d/1oE93V3SgOGWJ4O1zeD1UmpeToa0ZiiO6LqRAmZBPFWM/edit 
@toelke

toelke:
To clarify something which I misrepresented during the meeting today: In proposal 1, the name does not come from the `NetworkAttachmentDefinition`, but *from the CNI configuration* which always exists -- so no problem when not running with multus!

toelke:
@dougbtv please note that I have added a reference to the migration toolkit we discussed yesterday.

doc/proposals/named-networks.md
```diff
@@ -0,0 +1,231 @@
...
+See
+- https://github.com/k8snetworkplumbingwg/whereabouts/pull/256
+- https://github.com/k8snetworkplumbingwg/whereabouts/issues/50#issuecomment-874040513
+
+#### Migration toolkit
+
+Should the decision be made that scheme 1 or scheme 2 are implemented, we are facing a backwards-compatibility issue: whereabouts would not find existing `IPPool`s as they are stored under the CIDR-derived name.
+We propose to supply a small program (ideally also as OCI image for use in kubernetes `Job`s) that migrates an existing cluster from "legacy whereabouts" to the new, named and namespaced scheme.
+
+This tool would do roughly the following:
```
dougbtv, line 212: Thanks Phillip! This looks great overall, and let's also have it attain the leader election lock so that it locks out whereabouts requests during this migration if possible.

doc/proposals/named-networks.md
```diff
@@ -0,0 +1,231 @@
...
+<hr>
+
+## Design
+
+The network configuration already has a field `name`:
+
+```json
+{
+      "cniVersion": "0.3.0",
+      "name": "whereaboutsexample",
```
dougbtv, line 39: I think there's some complexity with "name" that it might make users have unexpected division of pools of IPs for existing net-attach-defs/CNI configs.

Given:

```
{
	"cniVersion": "0.3.0",
	"name": "macvlan-red",
	"type": "macvlan",
	"master": "eth0",
	"mode": "bridge",
	"ipam": {
		"type": "whereabouts",
		"range": "192.168.2.225/28"
	}
}

{
	"cniVersion": "0.3.0",
	"name": "macvlan-blue",
	"type": "macvlan",
	"master": "eth1",
	"mode": "bridge",
	"ipam": {
		"type": "whereabouts",
		"range": "192.168.2.225/28"
	}
}
```

Users would expect  `192.168.2.225/28` to be taken from the same IP pool across two configurations, the way it exists today.

Reusing the name would cause users to have to change their net-attach-defs.

If we add a new "name" type field to the whereabouts configuration in the IPAM section, then users would have to specifically add it, and... If it's not provided we could draw from a "default" pool. Such as:

```
{
	"cniVersion": "0.3.0",
	"name": "macvlan-color",
	"type": "macvlan",
	"master": "eth1",
	"mode": "bridge",
	"ipam": {
	    "pool_name": "whereabouts-blue-pool",
		"type": "whereabouts",
		"range": "192.168.2.225/28"
	}
}
```

dougbtv:
I think we make sure `enable_overlapping_ranges` option is disabled by default, and possibly think about deprecating that functionality. But yeah, maybe as a step one with this proposal, let's change to disabled by default. Thanks Phillips.

toelke:
How does `enable_overlapping_ranges` fit into this?

doc/proposals/named-networks.md
dougbtv: Maybe we need kind of a combination of both... Like, ideally not having the generated name is overall better.

But, I kind of don't want to break behavior that people might expect with existing CRDs.

Maybe if no name is specified, then we generate it. But, we honor the name when used. 

I'm still thinking it through -- Ayush also brought up that it's possible this could also cause confusion for users who forget to name a range, and then those pools get divided.

Ayush21298: How about implementing a comprehensive comparator for IP pools, which works on a combination of `pool_name` and `range` ?

If the user does not define the `pool_name`, we can add a wildcard for it. In the cases when the _wildcard_exists, we only use the `range` for comparison.

toelke: Instead of using `name` we can also be more explicit and call it something like `addressCollisionDomain` with a default of `""` -- that would also make the code more readable as the names of the `IPPools` would "just" be `<CIDR>-<addressCollisionDomain>`.

doc/proposals/named-networks.md
```diff
@@ -0,0 +1,231 @@
+# Support for named networks
+
+# Table of contents
+
+- [Introduction](#introduction)
+  - [Goal](#goal-of-this-proposal)
+- [Design](#design)
+  - [Changes in IPAM Config](#changes-in-ipam-config)
```
maiqueb, line 8: broken link

maiqueb:
Cool, this PR also worked as a side effect to prove #304 worked :)
The tests were skipped, as can be seen in the CI jobs:
- https://github.com/k8snetworkplumbingwg/whereabouts/actions/runs/4356187198/jobs/7613844290
- https://github.com/k8snetworkplumbingwg/whereabouts/actions/runs/4356187198/jobs/7613843736

dougbtv (approved):
LGTM! Thanks for the patience with the design phase. Let's go for it!

PR #261, opened 2022-08-08
controller: request more memory to prevent runtime OOM kills

### LINKED ISSUE #260: [BUG] OOM on ip-control-loop
**Describe the bug**
with k8s 1.22 I'm able to see/get more and more issues like
Memory cgroup out of memory: Killed process 643039 (ip-control-loop) total-vm:742792kB, anon-rss:43452kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:240kB oom_score_adj:-997

this error does not seem all the time and I'm also not sure about the impact of that on the PODS.
there is something I can check to understand from where it comes?

**Environment**:

    Limits:
      cpu:     100m
      memory:  50Mi
    Requests:
      cpu:     100m
      memory:  50Mi

- Whereabouts version : 
    Image:      ghcr.io/k8snetworkplumbingwg/whereabouts:latest-amd64
can i get the specific version somehow?

maiqueb:
We had forgotten to request more memory when we added the controller,
which furthermore now also runs the `ip-reconciler` cron.

None.

### DIFF
--- a/doc/crds/daemonset-install.yaml
+++ b/doc/crds/daemonset-install.yaml
@@ -110,10 +110,10 @@ spec:
         resources:
           requests:
             cpu: "100m"
-            memory: "50Mi"
+            memory: "100Mi"
           limits:
             cpu: "100m"
-            memory: "50Mi"
+            memory: "200Mi"
         securityContext:
           privileged: true
         volumeMounts:


dougbtv (approved):
I removed the fix for #260 -- I think we might want to revisit it a bit more. But this is indeed a perfect first step, thanks!

Issue #269, opened 2022-09-09
[BUG] cronJob error when installing cluster in k8s v1.25.0


subhamkrai:
**Describe the bug**
We are using these YAML
```
kubectl apply \
    -f https://raw.githubusercontent.com/k8snetworkplumbingwg/whereabouts/v0.5.3/doc/crds/daemonset-install.yaml \
    -f https://raw.githubusercontent.com/k8snetworkplumbingwg/whereabouts/v0.5.3/doc/crds/ip-reconciler-job.yaml \
    -f https://github.com/k8snetworkplumbingwg/whereabouts/raw/v0.5.3/doc/crds/whereabouts.cni.cncf.io_ippools.yaml \
    -f https://github.com/k8snetworkplumbingwg/whereabouts/raw/v0.5.3/doc/crds/whereabouts.cni.cncf.io_overlappingrangeipreservations.yaml
```
 to install multus and configure the network in minikube single node cluster for our CI testing([here are details steps](https://github.com/rook/rook/blob/master/tests/scripts/github-action-helper.sh#L431)). We can't upgrade our CI to use latest k8s version due error
```
error: resource mapping not found for name: "ip-reconciler" namespace: "kube-system" from "https://raw.githubusercontent.com/k8snetworkplumbingwg/whereabouts/v0.5.3/doc/crds/ip-reconciler-job.yaml": no matches for kind "CronJob" in version "batch/v1beta1"
```

We are using v0.5.3 and I see in latest release `/ip-reconciler-job.yaml` file has been removed. So we need some help getting our CI up and running.

**Expected behavior**
Installing whereabouts should go smooth in k8s v1.25.0

**To Reproduce**
1. use whereabouts v0.5.3 to install in k8s v1.25.0

**Environment**:
- Whereabouts version : v0.5.3
- Kubernetes version (use `kubectl version`): v1.25.0
- Network-attachment-definition: calico
- Whereabouts configuration (on the host): N/A
- OS (e.g. from /etc/os-release): ubuntu(github action)
- Kernel (e.g. `uname -a`): N/A
- Others: N/A

**Additional info / context**
CI link in case want more details https://github.com/rook/rook/runs/8242985791?check_suite_focus=true#step:7:42

Issue #178, opened 2022-01-19
k8s-1.25 - Whereabouts is using deprecated k8s API - CronJob


maiqueb:
Whereabouts is currently using some deprecated endpoints of the Kubernetes API; these logs appear when whereabout is deployed:
```
...
Warning: batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob
Warning: spec.template.spec.nodeSelector[beta.kubernetes.io/arch]: deprecated since v1.14; use "kubernetes.io/arch" instead
...
```

At least the `nodeSelector` can be updated right now - the `CronJob` probably can happen only later on, around k8s-1.24.

maiqueb:
Same as #269 .

PR #270, opened 2022-09-22
Make ip-reconciler job compatible with Kubernetes v1.25

e0ne:
CronJob API version is stable now, so we need to specify  `batch/v1` for it.

Closes: #178
Closes: #269

### DIFF
--- a/hack/e2e-setup-kind-cluster.sh
+++ b/hack/e2e-setup-kind-cluster.sh
@@ -15,8 +15,8 @@ done
 
 HERE="$(dirname "$(readlink --canonicalize ${BASH_SOURCE[0]})")"
 ROOT="$(readlink --canonicalize "$HERE/..")"
-MULTUS_DAEMONSET_URL="https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/master/deployments/multus-daemonset.yml"
-CNIS_DAEMONSET_URL="https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/master/e2e/cni-install.yml"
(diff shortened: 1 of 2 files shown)


hack/e2e-setup-kind-cluster.sh
```diff
@@ -15,8 +15,8 @@ done
 
 HERE="$(dirname "$(readlink --canonicalize ${BASH_SOURCE[0]})")"
 ROOT="$(readlink --canonicalize "$HERE/..")"
-MULTUS_DAEMONSET_URL="https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/master/deployments/multus-daemonset.yml"
-CNIS_DAEMONSET_URL="https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/master/e2e/cni-install.yml"
+MULTUS_DAEMONSET_URL="https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/v3.9.1/deployments/multus-daemonset.yml"
+CNIS_DAEMONSET_URL="https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/v3.9.1/e2e/cni-install.yml"
```
maiqueb, line 19: What error do you currently see in CI ? 

I worry a bit about this change, because it implies we have to track multus releases ... Something which I would rather not have to do.

@e0ne

maiqueb: @e0ne I only now realized you're pushing this directly to `release-v0.5` branch; wouldn't it make more sense to merge it to master/main branch, and then backport it into the release branch ?...

If we just do this, the main branch would still have a deprecated attribute.

maiqueb: @e0ne would you please update change this PR so it merges in the master branch ? We can backport to v0.5.0 afterwards.

PR #271, opened 2022-09-26
initial

kmlebedev:

migrations to golang 1.19


kmlebedev:
@dougbtv done. https://github.com/kmlebedev/whereabouts/actions/runs/3188703706

.github/workflows/build.yml
```diff
@@ -4,7 +4,7 @@ jobs:
   build:
     strategy:
       matrix:
-        go-version: [1.17.x, 1.18.x
```
maiqueb, line 7: shouldn't we also check 1.18 ? It's a [supported](https://endoflife.date/go) golang version.

e2e/client/ippool.go
```diff
@@ -16,8 +14,7 @@ import (
 func isIPPoolAllocationsEmpty(k8sIPAM *kubeClient.KubernetesIPAM, ipPoolName string) wait.ConditionFunc {
 	return func() (bool, error) {
 		ipPool, err := k8sIPAM.GetIPPool(context.Background(), ipPoolName)
-		noPoolError := fmt.Errorf("k8s pool initialized")
-		if errors.As(err, &noPoolError) {
+		if err.Error() == kubeClient.ErrKMsg8sPoolInit {
```
maiqueb: what's the advantage of comparing strings vs using the golang errors API ? I mean, why don't we just fix the `errors.As` call ? 

According to golang's release notes, go vet now just "reports when [errors.As](https://tip.golang.org/pkg/errors/#As) is called with a second argument of type *error, a common mistake."

pkg/storage/kubernetes/ipam.go
```diff
@@ -28,6 +28,10 @@ import (
 	"gomodules.xyz/jsonpatch/v2"
 )
 
+const (
+	ErrKMsg8sPoolInit = "k8s pool initialized"
```
maiqueb: typo.

e2e/client/ippool.go
```diff
@@ -14,7 +14,7 @@ import (
 func isIPPoolAllocationsEmpty(k8sIPAM *kubeClient.KubernetesIPAM, ipPoolName string) wait.ConditionFunc {
 	return func() (bool, error) {
 		ipPool, err := k8sIPAM.GetIPPool(context.Background(), ipPoolName)
-		if err.Error() == kubeClient.ErrMsgK8sPoolInit {
+		if err != nil && err.Error() == kubeClient.ErrMsgK8sPoolInit {
```
maiqueb, line 17: this wouldn't be needed if you just used the errors API - i.e. `errors.As` .

maiqueb (changes requested):
It's confusing to see the PR structured as is: multiple times a commit is just fixing something which was broken in a previous commit.

This makes the reviewers life harder, and renders tools like `git bisect` useless if this happens to be merged.

Could you please squash everything into a single commit - or structure the commits in a better way ? 

Finally, I wonder if we should jump to 1.19 *now*. I personally would only do it once 1.18 goes EOL - meaning, I think the project's go.mod should use the oldest supported version of golang.

.github/workflows/build.yml
kmlebedev: > shouldn't we also check 1.18 ? It's a [supported](https://endoflife.date/go) golang version.

I don't know how it will work when go.mod is explicitly set to version 1.19

e2e/client/ippool.go
kmlebedev: > this wouldn't be needed if you just used the errors API - i.e. `errors.As` .

I corrected, only what fell with errors and did not pass tests without optimizations and logic replacement
```
Run go vet --tags=test ./...
  go vet --tags=test ./...
  shell: /usr/bin/bash -e {0}
  env:
    GOROOT: /opt/hostedtoolcache/go/1.19.1/x64
# github.com/k8snetworkplumbingwg/whereabouts/e[2](https://github.com/kmlebedev/whereabouts/actions/runs/3184060412/jobs/5192029033#step:9:2)e/client
Error: e2e/client/ippool.go:20:[6](https://github.com/kmlebedev/whereabouts/actions/runs/3184060412/jobs/5192029033#step:9:7): second argument to errors.As should not be *error
Error: Process completed with exit code 2.
```

https://github.com/kmlebedev/whereabouts/actions/runs/3184060412/jobs/5192029033

e2e/client/ippool.go
kmlebedev: > This makes the reviewers life harder, and renders tools like `git bisect` useless if this happens to be merged.

I agree, it took a while to run all the tests locally and fix.

e2e/client/ippool.go
maiqueb: Yes, but that happens because we're currently passing a pointer to an error instead of the error itself. 

I am nearly sure if you do the following `go vet` will **not** complain:
```golang
		noPoolError := fmt.Errorf("k8s pool initialized")
		if errors.As(err, noPoolError) {
		...
```
Above I just removed the `&` from `noPoolError`.

dougbtv:
@kmlebedev any chance you can rebase? We can take another look at moving to 1.19 thanks for the PR

PR #277, opened 2022-10-28
Fix typo in Github workflow

antoninbas:
Signed-off-by: Antonin Bas <abas@vmware.com>


maiqueb (approved):
Good catch. I don't understand how did this slip through.

PR #296, opened 2023-01-18
Return existing reservation if podRef matches

xagent003:


nicklesimba (commented):
Mostly looks good and thanks for the detailed breakdown for the PR description. Just have one question...also will try to bring this up to others for further review.

s1061123 (commented):
Thank you for the PR. The diff makes sense to me. Could you please add unit-test in allocate_test.go for allocate.go changes, if you don't mind it?

moradiyaashish:
Till fix is available, is there any workaround available to recover from this situation ?

caribbeantiger:
> @xagent003 hey o/
> 
> This slipped me by; could you rebase ? Do you need help sorting out the unit tests ?

if @xagent003  doesn't mind , i can open a new merge request with this fix, as we are running into this issue often while testing  high availability and node crash scenarios.

@maiqueb  is that okay?

Ritwik037:
@dougbtv do you have any plan merging this PR because we are facing this issue with version 0.6.2 also and @xagent003 can you please confirm in which whereabouts version you made this changes and is it compatible with kubernetes 1.25

Ritwik037:
@maiqueb we are facing the issue that during the rolling upgrade of the kubernetes whenever a pods shift to a new worker node it got stuck the whereabout is not able to assign the ips to we tried the latest version 0.6.2 also but it didn't make any difference

PR #304, opened 2023-03-07
ci: replace a list of files for which tests must be run by an ignore list

maiqueb:
Saves resources, improves developer workflow since there's no point in running e2e tests when the documentation
is updated.

The pictures in the `doc/...` folder are also ignored.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,17 +20,9 @@ jobs:
       id: changed-files
       uses: tj-actions/changed-files@v35
       with:
-        files: |
-          .github
-          cmd
-          pkg
-          Dockerfile*
-          e2e
-          go.mod
-          go.sum
-          hack
-          script
-          doc/crds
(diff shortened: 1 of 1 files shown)


maiqueb:
Seems it ignored the Makefile changes as expected:
- https://github.com/k8snetworkplumbingwg/whereabouts/actions/runs/4355473768/jobs/7612142190
- https://github.com/k8snetworkplumbingwg/whereabouts/actions/runs/4355473768/jobs/7612141999

Going to update the PR to run the test when the Makefile is updated ...

maiqueb:
OK, it ran the tests using the ignore list in https://github.com/k8snetworkplumbingwg/whereabouts/actions/runs/4355903423/jobs/7613154521

Now removed the Makefile changes, and the tests should be executed - since we're touching the github actions configuration.

dougbtv (approved):
nice one, thanks Miguel

PR #309, opened 2023-03-15
Use downward API to pass current spec.nodeName to pod

andreaskaris:
The podInformerFactory uses filter key spec.nodeName to filter the pods that it should monitor. Up until now, this filter was set to the value of HOSTNAME. However, this is not reliable, as spec.nodeName can be overridden in kubernetes with --hostname-override and thus HOSTNAME and spec.nodeName do now necessarily always match. Instead, rely on a new custom environment variable NODENAME which is populated by the downward API.

Fixes https://issues.redhat.com/browse/OCPBUGS-10364

### DIFF
--- a/cmd/controlloop/controlloop.go
+++ b/cmd/controlloop/controlloop.go
@@ -9,9 +9,6 @@ import (
 
 	gocron "github.com/go-co-op/gocron"
 	corev1 "k8s.io/api/core/v1"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
-	v1coreinformerfactory "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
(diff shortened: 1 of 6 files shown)


cmd/controlloop/controlloop.go
```diff
@@ -180,3 +194,16 @@ func cronExpressionFromFlatFile() string {
 	}
 	return flatipam.IPAM.ReconcilerCronExpression
 }
+
+// nodeExists will return true if a node with the provided nodeName is found, false otherwise. The returned error will
+// be nil for NotFound errors, otherwise the error from the Nodes().Get() method will be returned.
+func nodeExists(k8sClientSet *kubernetes.Clientset, nodeName string) (bool, error) {
```
maiqueb, line 200: I think this signature is confusing.

Can't we just return an error and have the client compare against nil for knowing if the node exists ?

cmd/controlloop/controlloop.go
```diff
@@ -132,14 +139,21 @@ func newPodController(stopChannel chan struct{}) (*controlloop.PodController, er
...
+
+	nodeName := os.Getenv(podControllerNodeNameEnvVariable)
+	logging.Debugf("Filtering pods with filter key '%s' and filter value '%s'", podControllerFilterKey, nodeName)
+	if nodeFound, err := nodeExists(k8sClientSet, nodeName); !nodeFound {
+		msg := fmt.Sprintf("WARNING: Could not find node with node name '%s'. The pod informer may not work correctly.",
+			nodeName)
+		if err != nil {
+			msg = fmt.Sprintf("%s Received an unexpected error message, err: %q", msg, err)
+		}
+		logging.Errorf(msg)
```
maiqueb, line 151: I think we should flat out return an error here, making the pod crash loop if we can't find the node for which we want to "monitor" the workloads for.

maiqueb (changes requested):
Conceptually, this makes total sense - this had slipped me by - full FQDNs will not work.

Thanks for catching. 

Could you return an error instead, and throw in a couple of unit tests for this ? 
... sorry for the lack of current tests in this package ...

maiqueb:
@andreaskaris for the unit tests, you'll have to change the signature of https://github.com/k8snetworkplumbingwg/whereabouts/blob/87e95f163a9284959daaaa42ad12335f5daa0de5/cmd/controlloop/controlloop.go#L109 allowing you to inject the client set into it. 

K8S has a mock client set where you can provision it with the objects it would be able to retrieve: that way, you can write a negative and a positive test.

cmd/controlloop/controlloop.go
andreaskaris: done

cmd/controlloop/controlloop.go
andreaskaris: done

andreaskaris:
I addressed the unit tests a bit differently via the newDummyPodController @maiqueb @dougbtv mainly because I had already started rewriting it before I read the comments here. Let me know what you think - if that's not good enough, I can dig into @maiqueb 's suggestion

dougbtv (approved):
I'm good with the approach on the unit tests

PR #311, opened 2023-03-21
add qemu setup for arm build

BlaineEXE:
The ARM image build complains of a qemu script not being present. It seems that qemu is required for `docker buildx` to work.

Fixes #


.github/workflows/image-push-master.yml
```diff
@@ -59,4 +59,24 @@ jobs:
           push: true
           tags: ghcr.io/${{ github.repository }}:latest-ocp
           file: Dockerfile.openshift
-          platforms: linux/arm64
```
BlaineEXE, line 62: Here I realized in my last PR I assumed this build was arm, but it is for `latest-ocp`, so this also corrects that.

Issue #313, opened 2023-03-29
[RFE] Optimize IP allocation at scale


adrianchiris:
**Is your feature request related to a problem? Please describe.**
When scheduling multiple pods with multiple secondary networks it may take a while time until whereabouts allocates IPs for all interfaces if too many pods spin up at the same time possibly hitting kubelet default 4 min limit to run pod sandbox[1].

This has been encountered in K8s clusters (128 worker nodes) running AI/ML Jobs spinning up 128 Pods at the same time
each pod has a total of 17 networks, 16 of those are sriov + whereabouts as IPAM (essentially the same secondary network specified 16 times).

[1] https://github.com/kubernetes/kubernetes/blob/c3e7eca7fd38454200819b60e58144d5727f1bbc/pkg/kubelet/cri/remote/remote_runtime.go#L163

**Describe the solution you'd like**
So, i performed some experiments:

I have decided to try this in a more compact and simple environment (master + 2 workers all baremetal) creating a simple deployment with 128 replicas, specifying 16 additional networks (macvlan + whereabouts)
and indeed i have hit kubelet timeout before deployment was ready and some pods entered restart.

splitting to 16 different networks with separate IP ranges sped up the process as we now have less data to retrieve from k8s API on each call (and less work iterating over this data). now i got the same deployment running in about 3:35min

My next step was then to split the global lease used by whereabouts to a per pool lease. implementing some POC level code, i then tried it out and it now took 1:23min for the deployment to be ready (all pods running)

as a reference i deployed on my setup the same deployment (128 pods) with just primary network, no whereabouts involved and it took 1:10min for the deployment to be ready (all pods running)

So essentially my solution for optimizing IP allocation at scale consists of two things

1. Recommendation for user to split ranges (separate IPPools)
2. use leader election with lease per pool

I will upload a POC code shortly for this approach

**Describe alternatives you've considered**
An Alternative we have discussed internally was to avoid using leader election at CNI level and drive IP allocation from a central place

A controller which will watch for pods and assign ip addressed for their networks via CRD (create either CR instance per pod or per pod and network)

CNI plugin would just GET the object and return IPs within it or retry if not set.

This is going to be a relatively large change in both the approach of how whereabouts assigns IP as well as code base.

**Additional context**

adrianchiris:
@maiqueb @dougbtv  thoughts on this one ?

maiqueb:
On paper, it makes sense @adrianchiris .

[...]

PR #314, opened 2023-03-29
[POC] lease lock per IP pool

adrianchiris:
This is a POC level code for changing lease/leaderElection usage by whereabouts from a global lease to a per pool lease.

This PR is related to #313 and is used as additional context on the potential changes proposed.

### DIFF
--- a/pkg/storage/kubernetes/ipam.go
+++ b/pkg/storage/kubernetes/ipam.go
@@ -143,6 +143,11 @@ func normalizeRange(ipRange string) string {
 	return normalized
 }
 
+// leaseLockName returns the name of the leaseLock to be used for leader election
+func leaseLockName(pi *PoolIdentifier) string {
+	return fmt.Sprintf("whereabouts-%s", IPPoolName(*pi))
+}
+
 func (i *KubernetesIPAM) getPool(ctx context.Context, name string, iprange string) (*whereaboutsv1alpha1.IPPool, error) {
 	ctxWithTimeout, cancel := context.WithTimeout(ctx, storage.RequestTimeout)
 	defer cancel()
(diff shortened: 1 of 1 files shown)


maiqueb:
I'd say this makes sense without looking in detail !

Please be patient while we sort this out. 

If you can, please rebase and ensure the tests are passing.

@xagent003 could you review ?

pkg/storage/kubernetes/ipam.go
```diff
@@ -522,15 +531,40 @@ func IPManagementKubernetesUpdate(ctx context.Context, mode int, ipam *Kubernete
...
+					// reserve IP in overlappingIPRangeAllocation
+					err = overlappingrangestore.UpdateOverlappingRangeAllocation(requestCtx, mode, newip.IP, containerID, podRef)
+					// TODO: We should probably return a specific type of error in case k8s returns error AlreadyExsits, then exit early in case of any other error
+					if err != nil {
+						logging.Errorf("Failed to update overlappingRangeAllocation (%s). Continuing loop. %v", newip.IP, err)
+						continue
+					}
+
+					defer func() {
+						if err != nil {
```
xagent003, line 543: I dont think this will ever be hit? It only happens if err != nil but all previous conditions check for err != nil and continue or return.

Is there a reason for this defer func() at all? It does exactly the same as what was previously done above

pkg/storage/kubernetes/ipam.go
```diff
@@ -557,14 +592,6 @@ func IPManagementKubernetesUpdate(ctx context.Context, mode int, ipam *Kubernete
 			break RETRYLOOP
 		}
 
-		if ipamConf.OverlappingRanges {
```
xagent003, line 560: Can you clarify why this change was made - the code is common but was added back tot he Allocate and Deallocate case statements. Now the same code is duplicated in each case. Did this fix some other issue or optimization?

xagent003:
Hi @maiqueb I reviewed. the lease lock change looks good, but I'd like to understand the other change for moving around the overlappingranges code.

BTW we tested out this code and it does appear to improve performance.

maiqueb:
> Hi @maiqueb I reviewed. the lease lock change looks good, but I'd like to understand the other change for moving around the overlappingranges code.

@adrianchiris could you split this change to another commit and explain its goal in the commit msg ? 

> 
> BTW we tested out this code and it does appear to improve performance.

Thank you @xagent003 . I really appreciate your help.

PR #318, opened 2023-03-31
sign into registry for pushing arm master build

BlaineEXE:

Fixes #

### DIFF
--- a/.github/workflows/image-push-master.yml
+++ b/.github/workflows/image-push-master.yml
@@ -73,6 +73,14 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
 
+      - name: Login to Container Registry
+        if: github.repository_owner == 'k8snetworkplumbingwg'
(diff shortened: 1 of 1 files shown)


maiqueb (changes requested):
I don't follow this PR: We don't push this target to the container registry:
https://github.com/k8snetworkplumbingwg/whereabouts/pull/318/files#diff-d53663318eb602777e8a186022119265331e230442aea8a09beafa1ed1cac749R88

What's the point of logging to the registry then ?... 

The real question IMO is **why** we're not pushing the ARM target to the container registry. I honestly do not know if that is intentional. Maybe @dougbtv knows.

Never-the-less, IMO, this change only makes sense if you also push the target to the registry.

BlaineEXE:
@maiqueb I just rebased on latest master, and the `push: true` is there now. It was a PR based on an older master, so it would have resulted in the image being pushed before.

maiqueb (approved):
Thank you.

Sorry I didn't react sooner.

Issue #226, opened 2022-05-22
arm64 docker image


cyclinder:
Hi teams!

Is there an official arm64 image? Thanks

jonkerj:
I saw there is a dockerfile for ARM64, albeit a separate one. I think the project would benefit from multiarch images, which is not that hard to do.

I'm willing to write a PR for this, but to be frank: the lack on feedback on issues/PRs in this project is not encouraging me. If the devs like this approach and are willing to review such a PR, please let me know.

maiqueb:
I think we would be interested in providing an arm64 img. 

@jonkerj you can work on it if you want.

jonkerj:
I've created PR #324, which I believe solves all this

PR #323, opened 2023-04-05
Provide error when no config was found

jonkerj:
The control loop currently segfaults when none of the config files was found. This commit makes it give an actual error, providing the user with useful feedback.

**Which issue(s) this PR fixes** 
If needed, I'll raise an issue, please let me know.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -256,11 +256,11 @@ func GetFlatIPAM(isControlLoop bool, IPAM *types.IPAMConfig, extraConfigPaths ..
 			}
 
 			foundflatfile = confpath
-			return flatipam, foundflatfile, err
+			return flatipam, foundflatfile, nil
 		}
 	}
-	var err error
-	return flatipam, foundflatfile, err
+
+	return flatipam, foundflatfile, NewConfigFileNotFoundError()
 }
 
 func handleEnvArgs(n *types.Net, numV6 int, numV4 int, args types.IPAMEnvArgs) (int, int, error) {
@@ -368,6 +368,16 @@ func NewInvalidPluginError(ipamType string) *InvalidPluginError {
 
 func (e *InvalidPluginError) Error() string {
 	return fmt.Sprintf("only interested in networks whose IPAM type is 'whereabouts'. This one was: %s", e.ipamType)
+}
+
+type ConfigFileNotFoundError struct{}
+
+func NewConfigFileNotFoundError() *ConfigFileNotFoundError {
+	return &ConfigFileNotFoundError{}
+}
+
+func (e *ConfigFileNotFoundError) Error() string {
+	return "config file not found"
 }
 
 func storageError() error {
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path/filepath"
 	"testing"
 
 	. "github.com/onsi/ginkgo"
@@ -18,6 +19,18 @@ func TestAllocate(t *testing.T) {
 }
 
 var _ = Describe("Allocation operations", func() {
+	var tmpDir string
+
+	BeforeEach(func() {
+		var err error
+		tmpDir, err = os.MkdirTemp("", "whereabouts")
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	AfterEach(func() {
+		Expect(os.RemoveAll(tmpDir)).To(Succeed())
+	})
+
 	It("can load a basic config", func() {
 
 		conf := `{
@@ -37,7 +50,10 @@ var _ = Describe("Allocation operations", func() {
         }
       }`
 
-		ipamconfig, _, err := LoadIPAMConfig([]byte(conf), "")
+		confPath := filepath.Join(tmpDir, "whereabouts.conf")
+		Expect(os.WriteFile(confPath, []byte(conf), 0755)).To(Succeed())
+
+		ipamconfig, _, err := LoadIPAMConfig([]byte(conf), "", confPath)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(ipamconfig.LogLevel).To(Equal("debug"))
 		Expect(ipamconfig.LogFile).To(Equal("/tmp/whereabouts.log"))
@@ -167,7 +183,10 @@ var _ = Describe("Allocation operations", func() {
         ]
     }`
 
-		ipamconfig, err := LoadIPAMConfiguration([]byte(conf), "")
+		confPath := filepath.Join(tmpDir, "whereabouts.conf")
+		Expect(os.WriteFile(confPath, []byte(conf), 0755)).To(Succeed())
+
+		ipamconfig, err := LoadIPAMConfiguration([]byte(conf), "", confPath)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(ipamconfig.LogLevel).To(Equal("debug"))
 		Expect(ipamconfig.LogFile).To(Equal("/tmp/whereabouts.log"))
@@ -214,7 +233,10 @@ var _ = Describe("Allocation operations", func() {
         }
       }`
 
-		ipamConfig, _, err := LoadIPAMConfig([]byte(conf), "")
+		confPath := filepath.Join(tmpDir, "whereabouts.conf")
+		Expect(os.WriteFile(confPath, []byte(conf), 0755)).To(Succeed())
+
+		ipamConfig, _, err := LoadIPAMConfig([]byte(conf), "", confPath)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(ipamConfig.IPRanges[0].Range).To(Equal("192.168.1.0/24"))
 		Expect(ipamConfig.IPRanges[0].RangeStart).To(Equal(net.ParseIP("192.168.1.5")))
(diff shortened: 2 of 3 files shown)


pkg/config/config.go
```diff
@@ -259,8 +260,8 @@ func GetFlatIPAM(isControlLoop bool, IPAM *types.IPAMConfig, extraConfigPaths ..
 			return flatipam, foundflatfile, err
 		}
 	}
-	var err error
-	return flatipam, foundflatfile, err
+
+	return flatipam, foundflatfile, errors.New("no config file was found")
```
maiqueb, line 264: thanks , good catch.

Would you mind also changing line [260](https://github.com/k8snetworkplumbingwg/whereabouts/pull/323/files#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5R260) to `return flatipam, foundflatfile, nil` ? I think that's more explicit.

maiqueb (approved):

maiqueb (commented):
Could you also add a couple of unit tests to https://github.com/k8snetworkplumbingwg/whereabouts/blob/master/pkg/config/config_test.go ?

Let me know if you need any help with it.

jonkerj:
Sure, if I can bother you to look at #226 😉 (unrelated to this issue, but I think it deserves some attention too)

jonkerj:
> Could you also add a couple of unit tests to https://github.com/k8snetworkplumbingwg/whereabouts/blob/master/pkg/config/config_test.go ?
> 
> Let me know if you need any help with it.

Well I tried to, but then I discovered that a lot (maybe all) tests in `config_test.go` are failing now, because `GetFlamIPAM` (which throws an error when no config files are present) is invoked from [`LoadIPAMConfig`](https://github.com/k8snetworkplumbingwg/whereabouts/blob/master/pkg/config/config.go#L56). I'm not really sure if `LoadIPAMConfig` needs a bit of refactoring to detect this type of error, or if the tests need rewriting, or maybe even something else.  

Do you have a hint for me?

maiqueb:
> > Could you also add a couple of unit tests to https://github.com/k8snetworkplumbingwg/whereabouts/blob/master/pkg/config/config_test.go ?
> > Let me know if you need any help with it.
> 
> Well I tried to, but then I discovered that a lot (maybe all) tests in `config_test.go` are failing now, because `GetFlamIPAM` (which throws an error when no config files are present) is invoked from [`LoadIPAMConfig`](https://github.com/k8snetworkplumbingwg/whereabouts/blob/master/pkg/config/config.go#L56). I'm not really sure if `LoadIPAMConfig` needs a bit of refactoring to detect this type of error, or if the tests need rewriting, or maybe even something else.
> 
> Do you have a hint for me?

Yeah, so now that you return an error when there's no config present (I agree this should happen), the tests should be re-written to ensure this configuration is available to the tests.

Let's make sure @dougbtv is on board with the conceptual change first if you don't mind.

I can guide you on adapting the tests later on. 

Stay tuned.

maiqueb:
FWIW, the changes you need to do in the tests are quite small.

Here's what you need to do:

```golang
       var tmpDir string // this must be accessible from the tests
 
       BeforeEach(func() {
               var err error
               tmpDir, err = os.MkdirTemp("", "whereabouts") // must create a tmp dir to store the configs
       })

       AfterEach(func() {
               Expect(os.RemoveAll(tmpDir)).To(Succeed()) // must clean-up the generated configs
       })

      It("can load a basic config", func() {
		conf := `` // each test will have a different conf; you don't have to change anything here

		confPath := fmt.Sprintf("%s/%s", tmpDir, "whereabouts.conf") 
		Expect(os.WriteFile(confPath, []byte(conf), 0755)).To(Succeed()) // must persist the conf in the file-system

		ipamconfig, _, err := LoadIPAMConfig([]byte(conf), "", confPath) // must pass the new conf-path to the func; it only knows about the default paths
		... // further expectations, each tests looks for different stuff
	})
```

dougbtv (approved):

maiqueb:
@jonkerj alright! @dougbtv 's on board with this change.

Please update the failing tests in a similar way to what I'm proposing in [this link](https://github.com/k8snetworkplumbingwg/whereabouts/pull/323#issuecomment-1519884297).

Also please add a unit test that **expects** an error to be thrown - because the config on disk is **not** available.

jonkerj:
Hi, sounds allright, I will look into that in a week or so.

jonkerj:
I've updated my PR with a commit that fixes those tests. All is green on my side. I am recycling the filename `whereabouts.conf` throughout the tests, because (if I understand everything correct) it gets reinitialized every test (`BeforeEach` et al) or gets overwritten anyways.

pkg/config/config_test.go
```diff
@@ -49,6 +67,11 @@ var _ = Describe("Allocation operations", func() {
 
 	})
 
+	It("throws an error when no flat-files are found", func() {
+		_, _, err := GetFlatIPAM(true, &types.IPAMConfig{}, "")
+		Expect(err).To(HaveOccurred())
```
maiqueb, line 72: please match against the specific error using the `MatchError` matcher.

pkg/config/config_test.go
```diff
@@ -49,6 +67,11 @@ var _ = Describe("Allocation operations", func() {
 
 	})
 
+	It("throws an error when no flat-files are found", func() {
+		_, _, err := GetFlatIPAM(true, &types.IPAMConfig{}, "")
```
maiqueb, line 71: please omit the last attribute; [GetFlatIPAM](https://github.com/k8snetworkplumbingwg/whereabouts/blob/e12772b9ef977269cfabe8c079748ebeb550bb0e/pkg/config/config.go#L225) has a variadic signature :

```suggestion
		_, _, err := GetFlatIPAM(true, &types.IPAMConfig{})
```

maiqueb (changes requested):
I only have two minor comments. Other than that, looks good.

pkg/config/config_test.go
jonkerj: I've created a new error type for this and the tests matches on this specifically (preferred that over prefix/string matching)

pkg/config/config_test.go
jonkerj: Sure, I missed that, oops

maiqueb:
You have plenty of failed unit tests in `cmd/whereabouts_test.go` . 

You need to take care of them.

jonkerj:
You are right, I need to fix the unit tests on way more places than I did. I only tested `github.com/k8snetworkplumbingwg/whereabouts/pkg/config`, previously, my bad.

jonkerj:
OK, we are almost there. There are still 5 unit tests failing in `cmd/whereabouts_test.go`, and I'm again unsure how to proceed. In one of the tests (`allocate and release addresses on ADD/DEL`), `ipamConfig(..)` is called, which in turn calls `config.LoadIPAMConfig(..)`, which now throws an error.

I think simply ignoring `ConfigFileNotFoundError` in `cmd/whereabouts_tests.go` around lines 1451-1454 would mimic the previous behavior (as no config found would be ignored anyways), right?

Also: I feel a bit awkward about both passing `byte[](conf)` *and* write it to a file and pass that to `LoadIPAMConfig` (feels like it's not really DRY). How do you think about that?

maiqueb:
> OK, we are almost there. There are still 5 unit tests failing in `cmd/whereabouts_test.go`, and I'm again unsure how to proceed. In one of the tests (`allocate and release addresses on ADD/DEL`), `ipamConfig(..)` is called, which in turn calls `config.LoadIPAMConfig(..)`, which now throws an error.
> 
> I think simply ignoring `ConfigFileNotFoundError` in `cmd/whereabouts_tests.go` around lines 1451-1454 would mimic the previous behavior (as no config found would be ignored anyways), right?

Yeah, that **was** the previous behavior. 

... but IMO, the goal of this PR is to get rid of the previous behavior and ensure an error is thrown when a config is not provided. This must be reflected in the unit tests.

IMO, you cannot just ignore the error; you should either do a proper setup of the tests,  **or** mock the *real* component, providing an implementation that does what you want. 

Maybe the later is simpler.

> 
> Also: I feel a bit awkward about both passing `byte[](conf)` _and_ write it to a file and pass that to `LoadIPAMConfig` (feels like it's not really DRY). How do you think about that?

The whole API could be improved to be honest ... 

If you choose to pursue that, please do it in a separate PR.

jonkerj:
> OK, we are almost there. There are still 5 unit tests failing in `cmd/whereabouts_test.go`, and I'm again unsure how to proceed. In one of the tests (`allocate and release addresses on ADD/DEL`), `ipamConfig(..)` is called, which in turn calls `config.LoadIPAMConfig(..)`, which now throws an error.
> 
> I think simply ignoring `ConfigFileNotFoundError` in `cmd/whereabouts_tests.go` around lines 1451-1454 would mimic the previous behavior (as no config found would be ignored anyways), right?

I did something else: I `os.WriteFile`'d the mashalled config to a tempfile, similar to your earlier suggestion. As the function prototype of `ipamConfig` does not allow for passing errors currently (that needs significant rewrite of the 5 involved tests), it `nil`s out when anything fails (which cause the test to fail). Not the prettiest, but all tests pass now.

maiqueb (approved):
Thank you for this change, and your willingness to jump through all the hurdles I kept setting for you.

jonkerj:
you're welcome!

PR #360, opened 2023-07-13
Denormalize IP name before checking if pod is alive

nicklesimba:
This commit solves the issue of normalized IP names being used for checking if a pod is alive, which resulted in overlappingrangeipreservations being falsely deleted by the ip-reconciler.

### DIFF
--- a/pkg/reconciler/iploop.go
+++ b/pkg/reconciler/iploop.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strings"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


maiqueb (commented):
Can you add a unit test for this ? 

It'll be easier to visualize the changes with it.

pkg/reconciler/iploop.go
```diff
@@ -164,9 +165,12 @@ func (rl *ReconcileLooper) findClusterWideIPReservations(ctx context.Context) er
 
 	for _, clusterWideIPReservation := range clusterWideIPReservations {
 		ip := clusterWideIPReservation.GetName()
+		// De-normalize the IP
```
s1061123, line 168: May need to add comment why it is required (e.g. to prevent deletion?) because calling isPodAlive() with de-normalized IP seems to be unique in the code.

maiqueb: Can you point at where we're normalizing the IP in this comment ?

nicklesimba: Rather than explain about "preventing deletion" (a specific bug) I am instead opting to explain why denormalization is correct. I have also pointed at where the IP is normalized in this comment:

 "In the UpdateOverlappingRangeAllocation function, the IP address is created with a "normalized" name to comply with the k8s api. We must denormalize here in order to properly look up the IP address in the regular format, which pods use."

dougbtv (approved):

PR #365, opened 2023-07-19
IterateForAssignment: Properly handle invalid syntax for exclude range

### LINKED ISSUE #364: [BUG] exclude range currently does not correctly handle invalid syntax
**Describe the bug**
Invalid syntax in an exclude range is currently ignored, and a nil pointer is added to excludeRanges, leading to a nil pointer exception in skipExludedSubnets
This also includes cases where the excludeRange is a single IP address

**Expected behavior**

**To Reproduce**
1.  See the unit tests for the submitted patch
2. 
3.

**Environment**:
- Whereabouts version : N/A
- Kubernetes version (use `kubectl version`): N/A
- Others: N/A

**Additional info / context**
Add any other information / context about the problem here.

andreaskaris:
Correctly handle errors from parsing exclude ranges to avoid issues with nil pointer exceptions.
Parse single IP addresses such as 192.168.123.10 or fe02::10 by and convert them to CIDRs by appending the correct 
prefix.

Fixes #364 

Fixes #

### DIFF
--- a/pkg/allocate/allocate.go
+++ b/pkg/allocate/allocate.go
@@ -102,10 +102,14 @@ func IterateForAssignment(ipnet net.IPNet, rangeStart net.IP, rangeEnd net.IP, r
 	for _, r := range reserveList {
 		reserved[r.IP.String()] = true
 	}
+
 	// Build excluded list, "192.168.2.229/30", "192.168.1.229/30".
 	excluded := []*net.IPNet{}
 	for _, v := range excludeRanges {
-		_, subnet, _ := net.ParseCIDR(v)
+		subnet, err := parseExcludedRange(v)
+		if err != nil {
+			return net.IP{}, reserveList, fmt.Errorf("could not parse exclude range, err: %q", err)
+		}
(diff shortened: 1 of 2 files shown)


pkg/allocate/allocate.go
```diff
@@ -144,3 +148,28 @@ func skipExcludedSubnets(ip net.IP, excluded []*net.IPNet) net.IP {
...
+	// we got earlier.
+	ip := net.ParseIP(s)
+	if ip == nil {
+		return nil, err
+	}
+	// If the address parses, check if it's IPv4 or IPv6 and add the correct prefix.
+	if ip.To4() != nil {
+		_, subnet, err = net.ParseCIDR(fmt.Sprintf("%s/32", s))
+	} else {
+		_, subnet, err = net.ParseCIDR(fmt.Sprintf("%s/128", s))
```
dougbtv, line 172: Yep, I'm into this approach of adding the suffix if it's a single IP, nice call

dougbtv (approved):

PR #317, opened 2023-03-31
Ip reconcile cron config via ENV

caribbeantiger:
**What this PR does / why we need it**: Support for custom IP Reconciler  cron Schedule via Environment variable

**Which issue(s) this PR fixes** : #316

PR #392, opened 2023-11-22
Extended support for customizing whereabouts ip-reconciler cron schedule

nicklesimba:
Added a configmap with "cron_schedule" key value pair. The value of cron_schedule (i.e. "30 4 * * *") will be put into a file of the same name on whereabouts ip-reconciler pods, such that when the reconciler is enabled, the file should appear on the pod.

Whereabouts will now search for the reconciler_cron_file, and if present, it will use the cron schedule from the file. If the file is not present a default value from the flatfile will be used instead.

### DIFF
--- a/cmd/controlloop/controlloop.go
+++ b/cmd/controlloop/controlloop.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"strings"
 	"time"
 
 	gocron "github.com/go-co-op/gocron"
@@ -28,8 +29,9 @@ import (
 )
 
 const (
-	allNamespaces  = ""
-	controllerName = "pod-ip-controlloop"
+	allNamespaces        = ""
+	controllerName       = "pod-ip-controlloop"
+	reconciler_cron_file = "cron-schedule/reconciler_cron_file"
 )
 
 const (
@@ -66,7 +68,7 @@ func main() {
 	defer networkController.Shutdown()
 
 	s := gocron.NewScheduler(time.UTC)
-	schedule := cronExpressionFromFlatFile()
+	schedule := determineCronExpression()
 
 	_, err = s.Cron(schedule).Do(func() { // user configurable cron expression in install-cni.sh
 		reconciler.ReconcileIPs(errorChan)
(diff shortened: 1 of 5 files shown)


maiqueb:
Hm I wonder if this isn't good enough for what you need: https://github.com/k8snetworkplumbingwg/whereabouts/pull/317

IIUC, the user would edit the existing ds - configuring the cron via env variable, and kill the ds pods.

nicklesimba:
> Hm I wonder if this isn't good enough for what you need: #317
> 
> IIUC, the user would edit the existing ds - configuring the cron via env variable, and kill the ds pods.

This PR provides support for Openshift users to edit cron schedule via configmap. I think this is a separate feature. 

See: https://github.com/k8snetworkplumbingwg/whereabouts/pull/392/files#diff-307d21cb211139c5c6d37b29f30d91ccd3d8b39ea8a338c984b560ec0769e3f1R160

nicklesimba:
After some investigation, I think this support might still be necessary. What Miguel linked allows users to change the cron schedule at install time, but these changes allow a user to update during runtime.

nicklesimba:
My understanding is that the whereabouts install-cni.sh script has the following definition:

https://github.com/k8snetworkplumbingwg/whereabouts/pull/392/files#diff-c8589450558444bb4def3f2633a8eb70db55df1c2a86366d265aeb768e3e625bR16

This makes it so that when installation happens, whatever environment variable is added by the configmap (even if it's correctly updated) will get overwritten by the above line. 

At least, this is the behavior I observed, and the explanation I gave is my hypothesis, but I might be misunderstanding installation order.

script/install-cni.sh
```diff
@@ -19,7 +19,7 @@ WHEREABOUTS_RECONCILER_CRON=${WHEREABOUTS_RECONCILER_CRON:-30 4 * * *}
 
 mkdir -p $CNI_CONF_DIR/whereabouts.d
 WHEREABOUTS_KUBECONFIG=$CNI_CONF_DIR/whereabouts.d/whereabouts.kubeconfig
-WHEREABOUTS_FLATFILE=$CNI_CONF_DIR/whereabouts.d/whereabouts.conf
+WHEREABOUTS_FLATFILE=$CNI_CONF_DIR/whereabouts.d/whereabouts.conf # Yuki~ Nikhil's note: imo we should remove "flatfile" from whereabouts vocabulary and call this "WHEREABOUTS_CONF_FILE" instead. Flatfile may be the format but it's confusing naming.
```
maiqueb, line 22: +1 on the naming.

maiqueb:
> My understanding is that the whereabouts install-cni.sh script has the following definition:
> 
> https://github.com/k8snetworkplumbingwg/whereabouts/pull/392/files#diff-c8589450558444bb4def3f2633a8eb70db55df1c2a86366d265aeb768e3e625bR16
> 
> This makes it so that when installation happens, whatever environment variable is added by the configmap (even if it's correctly updated) will get overwritten by the above line.
> 
> At least, this is the behavior I observed, and the explanation I gave is my hypothesis, but I might be misunderstanding installation order.

[This line](https://github.com/k8snetworkplumbingwg/whereabouts/blob/0d7d2be59005359f33ec2d67b0ab004f9c59f291/script/install-cni.sh#L16) is setting a default: i.e. it will define `WHEREABOUTS_RECONCILER_CRON` with the value of the `WHEREABOUTS_RECONCILER_CRON`  environment value, or with "30 4 * * *" is the env variable is not defined.

But if you're saying it does not work ... That is probably something else we need to figure out. I can't say why it does not work as expected since I don't know.

doc/crds/daemonset-install.yaml
```diff
@@ -130,10 +143,16 @@ spec:
           mountPath: /host/opt/cni/bin
         - name: cni-net-dir
           mountPath: /host/etc/cni/net.d
+        - name: cron-scheduler-configmap
+            mountPath: /cron-schedule
```
maiqueb, line 147: wrong indentation. I guess this is why the daemonset does not start in e2e tests

cmd/controlloop/controlloop.go
```diff
@@ -66,7 +68,7 @@ func main() {
 	defer networkController.Shutdown()
 
 	s := gocron.NewScheduler(time.UTC)
-	schedule := cronExpressionFromFlatFile()
+	schedule := determineCronExpression()
```
maiqueb, line 71: hm, I think you still need to react to updates to the cron expression; meaning, this will only occur when the process starts.

Either you use an higher lib - like [fsnotify](https://github.com/fsnotify/fsnotify) - or implement something like that using a goroutine that checks if the file's contents have changed, and if so, restart the scheduler.

I guess even exiting the process could be good enough (but worse than reloading the config), since the pod would restart.

PR #399, opened 2023-12-20
Rebases #180: ip-reconciler: Use ContainerID instead of PodRef

dougbtv:
This is a (somewhat manual) rebase of #180 -- 

It looks like it's working in my testing of it, but... I initially had some **incorrect keys deleted only  sometimes**  when I was first testing it, and getting the wrong IPs deleted? 

So I was getting failures like:

```
Whereabouts IP reconciler
/home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/reconciler/ip_test.go:34
  reconciling an IP pool with multiple pods attached
  /home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/reconciler/ip_test.go:98
    each with IP from the same IPPool
    /home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/reconciler/ip_test.go:122
      reconciling the IPPool
      /home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/reconciler/ip_test.go:139
        should report the dead pod's IP address as deleted [It]
        /home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/reconciler/ip_test.go:146

        Expected
            <[]net.IP | len:1, cap:1>: [
                [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 10, 10, 10, 2],
            ]
        to equal
            <[]net.IP | len:1, cap:1>: [
                [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 10, 10, 10, 1],
            ]

        /home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/reconciler/ip_test.go:150
```

However, with whatever finagling I was doing, those have gone away... Or so I think.

So I've been testing it in a loop to repeat it a bunch of times, like:

```bash
#!/bin/bash

for i in {1..20}; do
    echo "Running test-go.sh - Attempt $i"
    ./hack/test-go.sh > /tmp/test_output_$i.txt 2>&1
    if [ $? -ne 0 ]; then
        echo "Test failed on attempt $i, output saved to /tmp/test_output_$i.txt"
    else
        rm /tmp/test_output_$i.txt
        echo "Test passed on attempt $i"
    fi
done
```

### DIFF
--- a/pkg/reconciler/ip_test.go
+++ b/pkg/reconciler/ip_test.go
@@ -8,6 +8,9 @@ import (
 	"strings"
 	"testing"
 
+	"math/rand"
+	"strconv"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
@@ -150,6 +153,7 @@ var _ = Describe("Whereabouts IP reconciler", func() {
 					Expect(deletedIPAddrs).To(Equal([]net.IP{net.ParseIP("10.10.10.1")}))
 				})
 
+				// !bang this can also be a problem area...
 				It("the IPPool should have only the IP reservation of the live pod", func() {
 					deletedIPAddrs, err := reconcileLooper.ReconcileIPPools(context.TODO())
 					Expect(err).NotTo(HaveOccurred())
@@ -163,7 +167,10 @@ var _ = Describe("Whereabouts IP reconciler", func() {
 							PodRef: fmt.Sprintf("%s/%s", namespace, pods[livePodIndex].Name),
 						},
 					}
-					Expect(poolAfterCleanup.Spec.Allocations).To(Equal(remainingAllocation))
+
+					for key, val := range poolAfterCleanup.Spec.Allocations {
+						Expect(val.PodRef).To(Equal(remainingAllocation[key].PodRef), "Mismatch at key: "+key)
+					}
 				})
 			})
 		})
@@ -457,7 +464,8 @@ func generateIPPoolSpec(ipRange string, namespace string, poolName string, podNa
 	allocations := map[string]v1alpha1.IPAllocation{}
 	for i, podName := range podNames {
 		allocations[fmt.Sprintf("%d", i+1)] = v1alpha1.IPAllocation{
-			PodRef: fmt.Sprintf("%s/%s", namespace, podName),
+			PodRef:      fmt.Sprintf("%s/%s", namespace, podName),
+			ContainerID: strconv.Itoa(rand.Intn(1000)),
 		}
 	}
 	return &v1alpha1.IPPool{
--- a/pkg/reconciler/ip_test.go
+++ b/pkg/reconciler/ip_test.go
@@ -146,6 +146,7 @@ var _ = Describe("Whereabouts IP reconciler", func() {
 				It("should report the dead pod's IP address as deleted", func() {
 					deletedIPAddrs, err := reconcileLooper.ReconcileIPPools(context.TODO())
 					Expect(err).NotTo(HaveOccurred())
+					// !bang this looks like a problem area...
 					Expect(deletedIPAddrs).To(Equal([]net.IP{net.ParseIP("10.10.10.1")}))
 				})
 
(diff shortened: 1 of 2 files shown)


s1061123:
I understand that the PR changes whereabout to change the identifier of allocation from podRef(Pod namespace/name) to contianer id. So how about to change reconciler and other code to match between Pod and allocation pool, i.e. garbageCollectPodIPs():pkg/controlloop/pod.go?

dougbtv:
Tomo not sure I'm following... This is to change the identifier during reconciliation. Do you mean to change the identifier to pod and allocation pool? I'm probably being dense

dougbtv:
I actually ran this 50 times and got 3 failures...

One I believe was a flake with a timeout starting the network controller, and then these two errors:

```
------------------------------
• Failure [0.001 seconds]
Whereabouts IP reconciler
/home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/reconciler/ip_test.go:37
  reconciling an IP pool with multiple pods attached
  /home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/reconciler/ip_test.go:101
    each with IP from the same IPPool
    /home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/reconciler/ip_test.go:125
      reconciling the IPPool
      /home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/reconciler/ip_test.go:142
        the IPPool should have only the IP reservation of the live pod [It]
        /home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/reconciler/ip_test.go:157

        Mismatch at key: 1
        Expected
            <string>: default/pod1
        to equal
            <string>: 

        /home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/reconciler/ip_test.go:172
```

and

```
------------------------------
• Failure [1.003 seconds]
IPControlLoop
/home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/controlloop/pod_controller_test.go:37
  a running pod on a node
  /home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/controlloop/pod_controller_test.go:61
    IPPool featuring an allocation for the pod
    /home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/controlloop/pod_controller_test.go:116
      the network attachment is available
      /home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/controlloop/pod_controller_test.go:127
        when the associated pod is deleted
        /home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/controlloop/pod_controller_test.go:160
          stale IP addresses are garbage collected [It]
          /home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/controlloop/pod_controller_test.go:165

          Timed out after 1.001s.
          the ip control loop should have removed this stale address
          Expected
              <map[string]v1alpha1.IPAllocation | len:1>: {
                  "0": {
                      ContainerID: "",
                      PodRef: "default/tiny-winy-pod",
                  },
              }
          to be empty

          /home/doug/codebase/src/github.com/dougbtv/whereabouts/pkg/controlloop/pod_controller_test.go:170
```

s1061123:
> This is to change the identifier during reconciliation. Do you mean to change the identifier to pod and allocation pool? I'm probably being dense
Should it be changed?

https://github.com/k8snetworkplumbingwg/whereabouts/blob/0d7d2be59005359f33ec2d67b0ab004f9c59f291/pkg/controlloop/pod.go#L225

dougbtv:
So, in a discussion with Tomo, he demonstrated that under crio, the containerid doesn't match the pod sandbox, so, this won't work under all runtimes. So we'll need a different approach for having a uniqueid to delete on for all container runtimes.

zshi-redhat:
> So, in a discussion with Tomo, he demonstrated that under crio, the containerid doesn't match the pod sandbox, so, this won't work under all runtimes. So we'll need a different approach for having a uniqueid to delete on for all container runtimes.

I discussed with Tomo, it seems the containerID and sandboxID is the same on openshift with crio (I tested it on 4.14 with this PR applied)

s1061123:
> I discussed with Tomo, it seems the containerID and sandboxID is the same on openshift with crio (I tested it on 4.14 with this PR applied)
Let me correct that. ContainerID and sandboxID is different in crio/OpenShift, but Zenghui test/verified that the PR fixes the issue.

pliurh:
@mlguerrero12 I proposed a new PR to address this issue. PTAL.

Issue #176, opened 2022-01-18
ip-reconciler cleans up incorrect IP, leads to duplicate IP


xagent003:
Discovered the issue via a duplicate IP scenario. This may be because ip-reconciler uses Pod name. With STS, pod names are always the same.

issue seen is:

1. Node A is rebooted / taken offline or some time. Let's say it had IP 151.2
2. STS Pod is terminated and re-created on a new node B, gets IP 151.1
3. ip-reconciler runs now, while node A is still rebooting (NotReady) or down. We see following in ip-reconciler logs:

```
2022-01-13T17:30:15Z [debug] pod reference pcrf/pcrf-sm-0 matches allocation; Allocation IP: 10.3.8.151; PodIPs: map[10.3.8.151:{} 10.4.2.2:{}]
2022-01-13T17:30:15Z [debug] pod reference pcrf/pcrf-sm-0 matches allocation; Allocation IP: 10.3.8.152; PodIPs: map[10.3.8.151:{} 10.4.2.2:{}]
2022-01-13T17:30:15Z [debug] pod ref pcrf/pcrf-sm-0 is not listed in the live pods list
```

First problem is that the podRef is a string. But for STS the names are not unique. Second problem is that is cleaned up the wrong IP, .151. This was the **new** IP in range that was assigned on node B. If it cleaned up an IP, it should have cleaned up the old one.

4. Now IPPool has following reservations (STS Pod highlighted in bold):

```
2022-01-13T17:30:15Z [debug] Going to update the reserve list to: [IP: 10.3.8.145 is reserved for pod: imaginemme/mme-sm-1 IP: 10.3.8.146 is reserved for pod: imaginemme/mme-sm-0 IP: 10.3.8.148 is reserved for pod: imaginemme/mme-sm-2 IP: 10.3.8.203 is reserved for pod: mvnr-mtcil1-appln-smf-smf1/smf-cdf-77fbd9d5bc-fbmdk **IP: 10.3.8.152 is reserved for pod: pcrf/pcrf-sm-0** IP: 10.3.8.153 is reserved for pod: pcrf/pcrf-sm-2 IP: 10.3.8.154 is reserved for pod: pcrf/pcrf-sm-1 IP: 10.3.8.202 is reserved for pod: mvnr-mtcil1-appln-smf-smf1/smf-cdf-77fbd9d5bc-qjk6k]
```

[...]

PR #433, opened 2024-03-04
Make podRef to be unique

pliurh:
We used podRef with format \<namespace>/<pod_name> to identify the IP allocation. However, in scenarios like replicaSet, a different pod can reuse the pod name. This patch changes the format of podRef to \<namespace>/<pod_name>:<pod_uid> to ensure each ip allocation will have a unique identifier. The legacy podRef format will be updated automatically during an upgrade.

### DIFF
--- a/cmd/controlloop/controlloop.go
+++ b/cmd/controlloop/controlloop.go
@@ -15,6 +15,7 @@ import (
 	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/tools/clientcmd"
 
 	nadclient "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/clientset/versioned"
 	nadinformers "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/informers/externalversions"
@@ -125,9 +126,19 @@ func handleSignals(stopChannel chan struct{}, signals ...os.Signal) {
 }
 
 func newPodController(stopChannel chan struct{}) (*controlloop.PodController, error) {
-	cfg, err := rest.InClusterConfig()
+	var cfg *rest.Config
+	var err error
+	cfg, err = rest.InClusterConfig()
 	if err != nil {
-		return nil, fmt.Errorf("failed to implicitly generate the kubeconfig: %w", err)
+		logging.Debugf("failed to generate the kubeconfig from service account: %v", err)
+		kubeConfigFile := os.Getenv("KUBECONFIG")
+		if kubeConfigFile == "" {
+			return nil, fmt.Errorf("KUBECONFIG environment variable not set")
+		}
+		cfg, err = clientcmd.BuildConfigFromFlags("", kubeConfigFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to generate kubeconfig from file %s: %v", kubeConfigFile, err)
+		}
 	}
 
 	k8sClientSet, err := kubernetes.NewForConfig(cfg)
--- a/cmd/controlloop/controlloop.go
+++ b/cmd/controlloop/controlloop.go
@@ -14,8 +14,8 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/record"
 
 	nadclient "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/clientset/versioned"
 	nadinformers "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/informers/externalversions"
@@ -39,6 +39,7 @@ const (
 	cronSchedulerCreationError
 	fileWatcherError
 	couldNotCreateConfigWatcherError
+	initialReconsileFailed
 )
 
 const (
@@ -78,6 +79,16 @@ func main() {
 		os.Exit(fileWatcherError)
 	}
 	defer watcher.Close()
+
+	// trigger one immediate reconcile before cron job start
+	go reconciler.ReconcileIPs(errorChan)
+	err = <-errorChan
+	if err == nil {
+		logging.Verbosef("initial reconciler success")
+	} else {
+		logging.Verbosef("initial reconciler failure: %s", err)
+		os.Exit(initialReconsileFailed)
+	}
 
 	reconcilerConfigWatcher, err := reconciler.NewConfigWatcher(
 		reconcilerCronConfiguration,
--- a/pkg/reconciler/iploop.go
+++ b/pkg/reconciler/iploop.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"regexp"
 	"strings"
 	"time"
 
@@ -54,17 +55,22 @@ func NewReconcileLooperWithClient(ctx context.Context, k8sClient *kubernetes.Cli
 		return nil, logging.Errorf("failed to retrieve all IP pools: %v", err)
 	}
 
+	looper := &ReconcileLooper{
+		k8sClient:      *k8sClient,
+		requestTimeout: timeout,
+	}
 	pods, err := k8sClient.ListPods(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	whereaboutsPodRefs := getPodRefsServedByWhereabouts(ipPools)
-	looper := &ReconcileLooper{
-		k8sClient:           *k8sClient,
-		liveWhereaboutsPods: indexPods(pods, whereaboutsPodRefs),
-		requestTimeout:      timeout,
+	// migrate the podRef format if needed
+	if err := looper.migrationPodRef(ctx, ipPools); err != nil {
+		return nil, logging.Errorf("failed to migrate podRef: %v", err)
 	}
+
+	whereaboutsPodRefs := getPodRefsServedByWhereabouts(ipPools)
+	looper.liveWhereaboutsPods = indexPods(pods, whereaboutsPodRefs)
 
 	if err := looper.findOrphanedIPsPerPool(ipPools); err != nil {
 		return nil, err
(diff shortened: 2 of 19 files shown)


pliurh:
fix #176

maiqueb:
Why have you chosen that format - `/<pod_name>:<pod_uid>` ? 

If you just need the pod ref to be unique, shouldn't we simply use the pod's UID ? 

Is the name part meant to help debugging or something like that ?

pliurh:
The new naming format shall be namespace/pod_name:pod_uid. And yes, it's meant for debugging.

cmd/controlloop/controlloop.go
```diff
@@ -125,9 +129,19 @@ func handleSignals(stopChannel chan struct{}, signals ...os.Signal) {
...
-		return nil, fmt.Errorf("failed to implicitly generate the kubeconfig: %w", err)
+		logging.Debugf("failed to generate the kubeconfig from service account: %v", err)
+		kubeConfigFile := os.Getenv("KUBECONFIG")
+		if kubeConfigFile == "" {
+			return nil, fmt.Errorf("KUBECONFIG environment variable not set")
+		}
+		cfg, err = clientcmd.BuildConfigFromFlags("", kubeConfigFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to generate kubeconfig from file %s: %v", kubeConfigFile, err)
+		}
```
dougbtv, line 144: thanks for the improved logging on error

pkg/reconciler/iploop.go
```diff
@@ -54,17 +55,22 @@ func NewReconcileLooperWithClient(ctx context.Context, k8sClient *kubernetes.Cli
...
 
-	pods, err := k8sClient.ListPods(ctx)
-	if err != nil {
+	looper := &ReconcileLooper{
+		k8sClient:      *k8sClient,
+		requestTimeout: timeout,
+	}
+
+	// migrate the podRef format if needed
+	if err := looper.migrationPodRef(ctx, ipPools); err != nil {
```
dougbtv, line 64: Ok, I think I get it how it triggers the migration, so here in the control loop, we check if the migration has happened and apply it here.

pkg/reconciler/iploop.go
```diff
@@ -54,17 +55,22 @@ func NewReconcileLooperWithClient(ctx context.Context, k8sClient *kubernetes.Cli
...
 		return nil, err
 	}
 
 	whereaboutsPodRefs := getPodRefsServedByWhereabouts(ipPools)
-	looper := &ReconcileLooper{
-		k8sClient:           *k8sClient,
-		liveWhereaboutsPods: indexPods(pods, whereaboutsPodRefs),
-		requestTimeout:      timeout,
+	pods, err := k8sClient.ListPods(ctx)
```
dougbtv, line 69: Can we limit this to just pods with `*/networks` annotations maybe? Unsure if this is premature optimization?

Something like:

```
    // Define label selector to filter pods with specific annotations
    labelSelector := metav1.LabelSelector{
        MatchExpressions: []metav1.LabelSelectorRequirement{
            {
                Key:      "k8s.v1.cni.cncf.io/networks",
                Operator: metav1.LabelSelectorOpExists,
            },
        },
    }

    // List pods matching the label selector
    pods, err := k8sClient.CoreV1().Pods("").List(ctx, metav1.ListOptions{
        LabelSelector: metav1.FormatLabelSelector(&labelSelector),
    })
```

pkg/reconciler/iploop_test.go
```diff
@@ -0,0 +1,108 @@
...
+	fakek8sclient "k8s.io/client-go/kubernetes/fake"
+
+	v1alpha1 "github.com/k8snetworkplumbingwg/whereabouts/pkg/api/whereabouts.cni.cncf.io/v1alpha1"
+	"github.com/k8snetworkplumbingwg/whereabouts/pkg/storage/kubernetes"
+
+	wbclient "github.com/k8snetworkplumbingwg/whereabouts/pkg/client/clientset/versioned"
+	fakewbclient "github.com/k8snetworkplumbingwg/whereabouts/pkg/client/clientset/versioned/fake"
+)
+
+var _ = FDescribe("MigrationPodRef", func() {
```
dougbtv, line 23: Thanks for the migration unit test.

pkg/reconciler/iploop.go
dougbtv: Sorry I had this wrong, this is abstracted. It's fine as is. Thanks!

pkg/reconciler/iploop_test.go
maiqueb: You need to remove the focus from these tests, otherwise, you'll highjack the CI to only run these tests.

maiqueb:
@dougbtv please allow me some time to review this PR. 

@pliurh you need to sort out the e2e and the unit tests. 

I'll keep peeking to see once they're good.

pkg/reconciler/iploop_test.go
pliurh: Good catch. I forgot it.

pliurh:
/unhold
@maiqueb @dougbtv I revised the PR. Please take another look.

cmd/controlloop/controlloop.go
```diff
@@ -125,9 +126,19 @@ func handleSignals(stopChannel chan struct{}, signals ...os.Signal) {
...
 
 func newPodController(stopChannel chan struct{}) (*controlloop.PodController, error) {
-	cfg, err := rest.InClusterConfig()
+	var cfg *rest.Config
+	var err error
+	cfg, err = rest.InClusterConfig()
 	if err != nil {
-		return nil, fmt.Errorf("failed to implicitly generate the kubeconfig: %w", err)
+		logging.Debugf("failed to generate the kubeconfig from service account: %v", err)
+		kubeConfigFile := os.Getenv("KUBECONFIG")
```
andreaskaris, line 145: both `newPodController` and `ReconcileIPs` are exclusively called from `controlloop.go:main()`.
I know that the operation doesn't cost anything, but it rubs me the wrong way to do `kubeConfigFile := os.Getenv("KUBECONFIG")` in each of them, the correct place imo is in `controlloop.go:main()` and then pass `kubeConfigFile` as an argument to both `newPodController` and `ReconcileIPs` . Perhaps that allows even for more optimizations

cmd/controlloop/controlloop.go
```diff
@@ -125,9 +126,19 @@ func handleSignals(stopChannel chan struct{}, signals ...os.Signal) {
 }
 
 func newPodController(stopChannel chan struct{}) (*controlloop.PodController, error) {
-	cfg, err := rest.InClusterConfig()
+	var cfg *rest.Config
+	var err error
+	cfg, err = rest.InClusterConfig()
```
andreaskaris, line 142: If this is run inside the cluster, but the KUBECONFIG env variable is set:

* ip.go will use NewReconcileLooperWithKubeconfig due to the if condition
* newPodController will use rest.InClusterConfig because in cluster rest.InClusterConfig will succeed

So your code can end up using the in cluster config for one, and the kubeconfig for the other

pkg/reconciler/iploop.go
```diff
@@ -170,8 +176,71 @@ func splitPodRef(podRef string) (string, string) {
...
+// 'Namespace/PodName:PodUID' for existing pools. This shall be a one-time job.
+func (rl ReconcileLooper) migrationPodRef(ctx context.Context, ipPools []storage.IPPool) error {
+	pattern := `^([^\s/]+)/([^\s/]+):([^\s/]+)$`
+
+	reservations, err := rl.k8sClient.ListOverlappingIPs(ctx)
+	if err != nil {
+		return logging.Errorf("failed to list overlappingrangeipreservations %v", err)
+	}
+	for _, ip := range reservations {
+		if regexp.MustCompile(pattern).MatchString(ip.Spec.PodRef) {
```
andreaskaris, line 207: nit - the regexp doesn't have to be compiled on every loop iteration:
```
const (
    podRefPattern = `^([^\s/]+)/([^\s/]+):([^\s/]+)$`
)

(...)

func (rl ReconcileLooper) migrationPodRef(ctx context.Context, ipPools []storage.IPPool) error {
    re := regexp.MustCompile(podRefPattern)
(...)
    for _, ip := range reservations {
		if re.MatchString(ip.Spec.PodRef) {
```

pkg/reconciler/iploop.go
```diff
@@ -170,8 +176,71 @@ func splitPodRef(podRef string) (string, string) {
...
+			return logging.Errorf("failed to update overlappingrangeipreservation %s: %v", ip.Spec.PodRef, err)
+		}
+	}
+
+	var ipReservations []types.IPReservation
+	for _, pool := range ipPools {
+		ipReservations = nil
+		needUpdate := false
+		for _, r := range pool.Allocations() {
+			if !regexp.MustCompile(pattern).MatchString(r.PodRef) {
```
andreaskaris, line 225: dito

pkg/reconciler/iploop_test.go
```diff
@@ -0,0 +1,108 @@
...
+		})
+
+		It("should update PodRef for overlappingrangeipreservations and pools", func() {
+			ipPools, err := reconcileLooper.k8sClient.ListIPPools(ctx)
+			Expect(err).NotTo(HaveOccurred())
+			err = reconcileLooper.migrationPodRef(ctx, ipPools)
+			Expect(err).NotTo(HaveOccurred())
+
+			found := false
+			pattern := `^([^\s/]+)/([^\s/]+):([^\s/]+)$`
```
andreaskaris, line 69: nit: reuse `podRefPattern` from `iploop.go` once that const is created, instead of redefining the pattern (less error prone):
```
const (
    podRefPattern = `^([^\s/]+)/([^\s/]+):([^\s/]+)$`
)
```

pkg/reconciler/iploop_test.go
```diff
@@ -0,0 +1,108 @@
...
+			// Allippools shall have been updated
+			ipPools, err = reconcileLooper.k8sClient.ListIPPools(ctx)
+			Expect(err).NotTo(HaveOccurred())
+			for _, pool := range ipPools {
+				for _, r := range pool.Allocations() {
+					if !regexp.MustCompile(pattern).MatchString(r.PodRef) {
```
andreaskaris, line 75: dito as earlier, compile outside of the loop, instead of compiling every single time

pkg/reconciler/iploop_test.go
```diff
@@ -0,0 +1,108 @@
...
+						found = true
+					}
+				}
+			}
+			Expect(found).To(BeFalse())
+			// All overlappingrangeipreservations shall have been updated
+			ips, err := reconcileLooper.k8sClient.ListOverlappingIPs(ctx)
+			Expect(err).NotTo(HaveOccurred())
+			for _, ip := range ips {
+				if !regexp.MustCompile(pattern).MatchString(ip.Spec.PodRef) {
```
andreaskaris, line 85: dito

pkg/storage/kubernetes/client.go
```diff
@@ -135,3 +137,18 @@ func (i *Client) DeleteOverlappingIP(ctx context.Context, clusterWideIP *whereab
...
+	defer cancel()
+
+	patchBytes, err := json.Marshal(map[string]interface{}{
+		"spec": clusterWideIP.Spec,
+	})
+	if err != nil {
+		return nil, logging.Errorf("Error marshaling overlappingrangeipreservations.whereabouts.cni.cncf.io spec: %v", err)
+	}
+
+	return i.client.WhereaboutsV1alpha1().OverlappingRangeIPReservations(clusterWideIP.GetNamespace()).Patch(
```
andreaskaris, line 152: question: does this have to be Patch, don't we have an Update() for that?  (I might be completely wrong)

andreaskaris: be it as it may, isn't https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict needed?

pliurh: The IP reconciler runs on multiple nodes, there will be racing conditions if using Update() in this case. However, using Patch() could cause too many API requests. We may need a leader election here.

I agree that we need to add a retry mechanism. This is a general problem in the IP reconciler, it doesn’t retry when reconciliation fails.

andreaskaris:
I've got a question ... will this actually work with Statefulsets and guarantee that the same IP address is always assigned to the same Statefulset pod? Because the uid between Statefulset pods changes:
```
 resourceVersion: ""
[akaris@workstation ~]$ oc get pods -o yaml | grep uid
      uid: b5a56769-43e0-4d8c-85bd-8a5f105986fb
    uid: 4ecd9b66-976e-4be3-84fa-cfe7939abfa9
[akaris@workstation ~]$ oc get pods -o yaml | grep uid
      uid: b5a56769-43e0-4d8c-85bd-8a5f105986fb
    uid: 4ecd9b66-976e-4be3-84fa-cfe7939abfa9
[akaris@workstation ~]$ oc delete pod tools-ipv4-0
pod "tools-ipv4-0" deleted
[akaris@workstation ~]$ 
[akaris@workstation ~]$ oc get pods
NAME           READY   STATUS              RESTARTS   AGE
tools-ipv4-0   0/1     ContainerCreating   0          3s
[akaris@workstation ~]$ oc get pods -o yaml | grep uid
      uid: b5a56769-43e0-4d8c-85bd-8a5f105986fb
    uid: f4dc518a-da66-4bad-832b-32394ea08299
[akaris@workstation ~]$ 
[akaris@workstation ~]$ 
[akaris@workstation ~]$ oc get pods -o yaml | grep uid
      uid: b5a56769-43e0-4d8c-85bd-8a5f105986fb
    uid: f4dc518a-da66-4bad-832b-32394ea08299
[akaris@workstation ~]$ 
[akaris@workstation ~]$ 
[akaris@workstation ~]$ oc get pods -o yaml | grep uid
      uid: b5a56769-43e0-4d8c-85bd-8a5f105986fb
    uid: f4dc518a-da66-4bad-832b-32394ea08299
```

I'm trying to understand when and why it would ever be a problem that an allocation is being reused (in the case of Statefulsets, isn't it even *desired* that an allocation is reused?) when the combination of namespace/name is actually a unique identifier for a pod

pliurh:
@andreaskaris I think you made a great point. Instead of addressing it from the IP reconciler's perspective, resolving it at the allocation stage would be more effective.

pliurh:
replaced by https://github.com/k8snetworkplumbingwg/whereabouts/pull/446

PR #441, opened 2024-03-15
Update k8s deps

manuelbuil:
We need to update:
* k8s.io/api
* k8s.io/apimachinery
* k8s.io/client-go
* k8s.io/code-generator
* k8s.io/kube-openapi
* sigs.k8s.io/controller-tools/cmd/controller-gen (not necessary but I saw that we were using a pretty old version)

but those updates fail because we are using deprecated functions:
```
Error: e2e/client/ippool.go:37:9: wait.PollImmediate is deprecated: This method does not return errors from context, use PollUntilContextTimeout. Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release.  (SA1019)
Error: e2e/client/pod.go:19:9: wait.PollImmediate is deprecated: This method does not return errors from context, use PollUntilContextTimeout. Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release.  (SA1019)
Error: e2e/client/pod.go:25:9: wait.PollImmediate is deprecated: This method does not return errors from context, use PollUntilContextTimeout. Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release.  (SA1019)
Error: e2e/client/replicaset.go:19:9: wait.PollImmediate is deprecated: This method does not return errors from context, use PollUntilContextTimeout. Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release.  (SA1019)
Error: e2e/client/replicaset.go:25:9: wait.PollImmediate is deprecated: This method does not return errors from context, use PollUntilContextTimeout. Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release.  (SA1019)
Error: e2e/client/statefulset.go:18:9: wait.PollImmediate is deprecated: This method does not return errors from context, use PollUntilContextTimeout. Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release.  (SA1019)
Error: e2e/client/statefulset.go:50:9: wait.PollImmediate is deprecated: This method does not return errors from context, use PollUntilContextTimeout. Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release.  (SA1019)
```
This PR replaces those deprecated functions with the recommended one: `PollUntilContextTimeout` and does the update on those k8s dependencies

Apart from that, given the `k8s.io/code-generator` update, which is required if we update the rest, we must add the line:
```
chmod +x vendor/k8s.io/code-generator/generate-internal-groups.sh
```
in the `Generate code` step of the workflow test. The problem is that we are using a deprecated script, we should move to kube_codegen.sh. As the warning says:

```
WARNING: generate-groups.sh is deprecated.
WARNING: Please use k8s.io/code-generator/kube_codegen.sh instead. 
```
I can try to do that in another PR (I think this one is big enough)

Fixes #


dougbtv (approved):
Thanks for the refactor for context changes. LGTM

PR #446, opened 2024-03-27
Return the allocated IP to a pod with the same namespace/name

pliurh:
For statefulset pods, it will reuse the pod names after a node reboot. Whereabouts shall always return the same IP address that has already been allocated, when the pods start.

This patch also changes the IP deallocation logic. Now, the allocation record is identified by podRef instead of containerID.

### DIFF
--- a/cmd/whereabouts_test.go
+++ b/cmd/whereabouts_test.go
@@ -62,13 +62,15 @@ func AllocateAndReleaseAddressesTest(ipVersion string, ipamConf *whereaboutstype
 		fakek8sclient.NewSimpleClientset(),
 		0)
 	for i := 0; i < len(expectedAddresses); i++ {
+		name := fmt.Sprintf("%s-%d", podName, i)
 		args := &skel.CmdArgs{
 			ContainerID: fmt.Sprintf("dummy-%d", i),
 			Netns:       nspath,
 			IfName:      ifname,
 			StdinData:   cniConf,
-			Args:        cniArgs(podNamespace, podName),
+			Args:        cniArgs(podNamespace, name),
 		}
+		ipamConf.PodName = name
 		client := mutateK8sIPAM(args.ContainerID, ipamConf, wbClient)
 
 		// Allocate the IP
@@ -76,7 +78,7 @@ func AllocateAndReleaseAddressesTest(ipVersion string, ipamConf *whereaboutstype
 			return cmdAdd(args, client, cniVersion)
 		})
 		Expect(err).NotTo(HaveOccurred())
-		Expect(strings.Index(string(raw), "\"version\":")).Should(BeNumerically(">", 0))
+		Expect(strings.Index(string(raw), "\"cniVersion\":")).Should(BeNumerically(">", 0))
 
 		result, err := current.GetResult(r)
 		Expect(err).NotTo(HaveOccurred())
@@ -132,6 +134,7 @@ func AllocateAndReleaseAddressesTest(ipVersion string, ipamConf *whereaboutstype
 var _ = Describe("Whereabouts operations", func() {
 	const (
 		podName      = "dummyPOD"
+		podNameSec   = "dummyPODSec"
 		podNamespace = "dummyNS"
 	)
 
@@ -952,16 +955,17 @@ var _ = Describe("Whereabouts operations", func() {
 				ipPool(ipamConf.IPRanges[0].Range, podNamespace, ipamConf.NetworkName)),
 			fakek8sclient.NewSimpleClientset(),
 			0)
-
 		var ipArgs []*skel.CmdArgs
 		// allocate 8 IPs (192.168.1.5 - 192.168.1.12); the entirety of the pool defined above
 		for i := 0; i < 8; i++ {
+			name := fmt.Sprintf("%s-%d", podName, i)
+			ipamConf.PodName = name
 			args := &skel.CmdArgs{
 				ContainerID: fmt.Sprintf("dummy-%d", i),
 				Netns:       nspath,
 				IfName:      ifname,
 				StdinData:   []byte(conf),
-				Args:        cniArgs(podNamespace, podName),
+				Args:        cniArgs(podNamespace, name),
 			}
 
 			k8sClient = mutateK8sIPAM(args.ContainerID, ipamConf, wbClient)
@@ -982,7 +986,7 @@ var _ = Describe("Whereabouts operations", func() {
 				}))
 			ipArgs = append(ipArgs, args)
 		}
-
+		ipamConf.PodName = podName
 		// assigning more IPs should result in error due to the defined range_start - range_end
 		args := &skel.CmdArgs{
 			ContainerID: "dummy-failure",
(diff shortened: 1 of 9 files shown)


pliurh:
fix https://github.com/k8snetworkplumbingwg/whereabouts/issues/176

pliurh:
@andreaskaris PTAL

cmd/whereabouts_test.go
```diff
@@ -62,21 +62,23 @@ func AllocateAndReleaseAddressesTest(ipVersion string, ipamConf *whereaboutstype
 		fakek8sclient.NewSimpleClientset(),
 		0)
 	for i := 0; i < len(expectedAddresses); i++ {
+		name := fmt.Sprintf(podName+"-%d", i)
```
andreaskaris, line 65:
```suggestion
name := fmt.Sprintf("%s-%d", i)
```

cmd/whereabouts_test.go
```diff
@@ -952,16 +955,17 @@ var _ = Describe("Whereabouts operations", func() {
 				ipPool(ipamConf.IPRanges[0].Range, podNamespace, ipamConf.NetworkName)),
 			fakek8sclient.NewSimpleClientset(),
 			0)
-
 		var ipArgs []*skel.CmdArgs
 		// allocate 8 IPs (192.168.1.5 - 192.168.1.12); the entirety of the pool defined above
 		for i := 0; i < 8; i++ {
+			name := fmt.Sprintf(podName+"-%d", i)
```
andreaskaris, line 961:
```suggestion
		name := fmt.Sprintf("%s-%d", podName, i)
```

pkg/allocate/allocate.go
```diff
@@ -129,7 +136,7 @@ func IterateForAssignment(ipnet net.IPNet, rangeStart net.IP, rangeEnd net.IP, r
 			continue
 		}
 		// Assign and reserve the IP and return.
-		logging.Debugf("Reserving IP: |%v|", ip.String()+" "+containerID)
+		logging.Debugf("Reserving IP: |%v|", ip.String()+" "+podRef+" "+containerID)
```
andreaskaris, line 139:
```suggestion
logging.Debugf("Reserving IP: |%s %s %s|", ip.String(), podRef, containerID)
```
... wondering if that yields the same output

pkg/allocate/allocate.go
```diff
@@ -101,7 +100,15 @@ func IterateForAssignment(ipnet net.IPNet, rangeStart net.IP, rangeEnd net.IP, r
 
 	// Build reserved map.
 	reserved := make(map[string]bool)
-	for _, r := range reserveList {
+	for i, r := range reserveList {
+		if r.PodRef == podRef {
+			logging.Debugf("find PodRef: %v with IP: %v allocated", podRef, r.IP)
```
andreaskaris, line 105: find -> found

pkg/storage/kubernetes/ipam.go
```diff
@@ -200,54 +200,63 @@ func (i *KubernetesIPAM) GetOverlappingRangeStore() (storage.OverlappingRangeSto
...
+				PodRef:      podRef,
+			}
+			_, err = c.client.WhereaboutsV1alpha1().OverlappingRangeIPReservations(c.namespace).Create(
+				ctx, clusteripres, metav1.CreateOptions{})
+		} else {
+			clusteripres.Spec = whereaboutsv1alpha1.OverlappingRangeIPReservationSpec{
+				ContainerID: containerID,
+				PodRef:      podRef,
+			}
+			_, err = c.client.WhereaboutsV1alpha1().OverlappingRangeIPReservations(c.namespace).Update(
```
andreaskaris, line 253: wrap in retry?

dougbtv:
I tried to surf around the history to figure out if there was a reason we didn't use podRef, but, I can't find any smoking gun. 

I think this approach of switching to base on podRef is an improvement, thanks.

Thanks Peng for the PR, and Andreas for the review, as well.

pkg/storage/kubernetes/ipam.go
pliurh: It will return an error and fail the CNI call. So kubelet will retry the CNI call. Additionally, we don't have the retry for other API calls either. Therefore, if we want to implement a retry logic, we want to address it comprehensively.

wwcd:
When a pod references multiple virtual function interfaces and all are allocated addresses, they will all be assigned the same address.

pliurh:
> When a pod references multiple virtual function interfaces and all are allocated addresses, they will be assigned the same address.

Yeah, it is a limitation when you have a pod with multiple interfaces using the same IP range. A workaround is to divide the IP range into sections and assign different ranges to each interface of the pod.

If we want to fix that, we need to update the API to store interface names. So that we can identify the IP allocation per interface. @dougbtv @andreaskaris @maiqueb WDYT

andreaskaris:
I'm trying to reproduce https://issues.redhat.com/browse/OCPBUGS-29648 or https://github.com/k8snetworkplumbingwg/whereabouts/issues/176 with upstream/master whereabouts on kind, to get a baseline for my testing.
I can reproduce OCPBUGS-29648 on OCP 4.14, but I cannot reproduce the same issue on kind; but I wouldn't exclude that I'm doing something wrong ... @pliurh could you repro the issues on upstream prior to applying the PR?

pliurh:
@andreaskaris I couldn't reproduce with a kind cluster either. I reproduced it with OCP.

pkg/storage/kubernetes/ipam.go
andreaskaris: yeah o.k. I misunderstood this anyway given that it's a CNI plugin it doesn't make a lot of sense

andreaskaris:
I just tested the `reboot --force` scenario from .. manually and everything works for me with the fix.

WRT attaching the same network 2x:
```
 25   template:                                                                                                             
 26     metadata:                                                                                                           
 27       annotations:                                                                                                      
 28         k8s.v1.cni.cncf.io/networks: test-ipv4,test-ipv4
```

I get:
```
[akaris@workstation whereabouts-cni ((dfaf5249...))]$ oc get pods tools-ipv4-0 --output=custom-columns="NAME:.metadata.name,NETWORK:.metadata.annotations.k8s\.v1\.cni\.cncf\.io\/network-status"
NAME           NETWORK
tools-ipv4-0   [{
    "name": "ovn-kubernetes",
    "interface": "eth0",
    "ips": [
        "10.128.0.101",
        "fd01:0:0:1::65"
    ],
    "mac": "0a:58:0a:80:00:65",
    "default": true,
    "dns": {}
},{
    "name": "test/test-ipv4",
    "interface": "net1",
    "ips": [
        "192.168.10.140"
    ],
    "mac": "ae:3d:0f:f3:28:4e",
    "dns": {}
},{
    "name": "test/test-ipv4",
    "interface": "net2",
    "ips": [
        "192.168.10.140"
    ],
    "mac": "22:3b:b7:71:d5:c9",
    "dns": {}
}]
[akaris@workstation whereabouts-cni ((dfaf5249...))]$ oc exec tools-ipv4-0 -- ip a | grep 192
    inet 192.168.10.140/24 brd 192.168.10.255 scope global net1
    inet 192.168.10.140/24 brd 192.168.10.255 scope global net2
```

cgoncalves:
LGTM

PR #458, opened 2024-04-17
Node Slice Fast IPAM

ivelichkovich:

improves performance with node slice mode

https://docs.google.com/document/d/1YlWfg3Omrk3bf6Ujj-s5wXlP6nYo4PZseA0bS6qmvkk/edit#heading=h.ehhncqtntm3t

Fixes #

This is a very very rough draft to help guide the design and discussion


cmd/nodeslicecontroller/node_slice_controller.go
```diff
@@ -0,0 +1,90 @@
+package main
```
ivelichkovich, line 1: the other controller runs per daemonset this will run as its own pod, in the future we can get the other controller to run in the same pod

cmd/whereabouts.go
```diff
@@ -76,7 +76,12 @@ func cmdAdd(args *skel.CmdArgs, client *kubernetes.KubernetesIPAM, cniVersion st
 	ctx, cancel := context.WithTimeout(context.Background(), types.AddTimeLimit)
 	defer cancel()
 
-	newips, err := kubernetes.IPManagement(ctx, types.Allocate, client.Config, client)
+	if client.Config.Range != "" && client.Config.NodeSliceFastRange != "" {
```
ivelichkovich, line 79: I guess error handling mixed configurations will depend on how we want to define the new IPAM Config fields

pkg/api/whereabouts.cni.cncf.io/v1alpha1/nodeslicepool_types.go
```diff
@@ -0,0 +1,52 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NodeSlicePoolSpec defines the desired state of NodeSlicePool
+type NodeSlicePoolSpec struct {
+	// Range is a RFC 4632/4291-style string that represents an IP address and prefix length in CIDR notation
+	// this refers to the entire range where the node is allocated a subset
+	Range string `json:"range"`
+
+	SliceSize string `json:"sliceSize"`
```
ivelichkovich, line 15: I'll add comments and all that once we have consensus on the design

pkg/node-controller/controller.go
```diff
@@ -0,0 +1,676 @@
...
+
+	nadInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: c.onNadUpdate,
+		UpdateFunc: func(old, cur interface{}) {
+			c.onNadUpdate(cur)
+		},
+		DeleteFunc: c.onNadDelete,
+	})
+
+	//nodeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```
ivelichkovich, line 152: do we want to reconcile if a user makes manual changes against our CRDS?

pkg/node-controller/controller.go
```diff
@@ -0,0 +1,676 @@
...
+	//TODO: error syncing 'default/igor-1': config file not found, requeuing
+	// can we make this work without a local config file
+	ipamConf, err := ipamConfiguration(nad, "")
+	if err != nil {
+		return err
+	}
+
+	// This is to support several NADs and interfaces on the same network
+	//TODO: validation that nad configs are compatible between different nads with same network-name
+	sliceName := name
```
ivelichkovich, line 393: this is an important feature for us, not sure if we could make it work nicely with overlapping ranges. We want to support many interfaces within the same network so many NADs will need to lock across the same node slices. This should also support "regular" NAD configs

this way it forces multiple nads to the same node slice allocations and then the same lease and IP Pool

pkg/node-controller/controller.go
```diff
@@ -0,0 +1,676 @@
...
+		"network-attachment-definition", klog.KRef(namespace, name))
+
+	currentNodeSlicePool, err := c.nodeSlicePoolLister.NodeSlicePools(namespace).Get(sliceName)
+	if err != nil {
+		logger.Info("node slice pool does not exist, creating")
+		if !errors.IsNotFound(err) {
+			return err
+		}
+		//Create
+		nodeslice := &v1alpha1.NodeSlicePool{
```
ivelichkovich, line 421: We probably want to clear out the node slices on delete (or when all nads using the node slice are deleted), do we also want to clear out the IPPools?

pkg/types/types.go
```diff
@@ -54,6 +54,8 @@ type IPAMConfig struct {
 	OmitRanges               []string             `json:"exclude,omitempty"`
 	DNS                      cnitypes.DNS         `json:"dns"`
 	Range                    string               `json:"range"`
+	NodeSliceFastRange       string               `json:"nodeSliceFastRange"`
```
ivelichkovich, line 57: need inputs on these fields, thinking we just reuse the existing Range field, add the node_slice_fast_range field and then we initially don't support rangestart/rangeend/overlapping_ranges in this mode

pkg/node-controller/controller.go
```diff
@@ -0,0 +1,676 @@
...
+	nadInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: c.onNadUpdate,
+		UpdateFunc: func(old, cur interface{}) {
+			c.onNadUpdate(cur)
+		},
+		DeleteFunc: c.onNadDelete,
+	})
+
+	//nodeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+	//	AddFunc: c.onNodeChange,
```
ivelichkovich, line 153: On node just re-reconcile the nads

pkg/node-controller/controller.go
```diff
@@ -0,0 +1,676 @@
...
+			}
+		} else {
+			//slices have not changed so only make sure all nodes are assigned
+			nodeslice := currentNodeSlicePool.DeepCopy()
+			allocations := currentNodeSlicePool.Status.Allocations
+			nodes, err := c.nodeLister.List(labels.Everything())
+			if err != nil {
+				return err
+			}
+			for _, node := range nodes {
```
ivelichkovich, line 510: Missed functions to remove node assignment

pkg/node-controller/controller.go
```diff
@@ -0,0 +1,676 @@
...
+	// MessageResourceExists is the message used for Events when a resource
+	// fails to sync due to a Deployment already existing
+	MessageResourceExists = "Resource %q already exists and is not managed by Foo"
+	// MessageResourceSynced is the message used for an Event fired when a Foo
+	// is synced successfully
+	MessageResourceSynced = "Foo synced successfully"
+)
+
+// Controller is the controller implementation for Foo resources
+type Controller struct {
```
ivelichkovich, line 60: I tried to follow a similar flow to KCM endpoints controller because it's a similar logical flow

pkg/storage/kubernetes/ipam.go
```diff
@@ -455,11 +483,36 @@ func IPManagement(ctx context.Context, mode int, ipamConf whereaboutstypes.IPAMC
...
+	}
+	return ipam.Config.NetworkName
+}
+
 // IPManagementKubernetesUpdate manages k8s updates
 func IPManagementKubernetesUpdate(ctx context.Context, mode int, ipam *KubernetesIPAM, ipamConf whereaboutstypes.IPAMConfig,
 	containerID string, podRef string) ([]net.IPNet, error) {
 	logging.Debugf("IPManagement -- mode: %v / containerID: %v / podRef: %v", mode, containerID, podRef)
-
+	logging.Debugf("test")
```
ivelichkovich, line 515: Again very WIP 😊

jingczhang:
NodeSlice CR is user to define and change...so hard to match the different runtime need of each node? not before the user is AI I guess. 
If we can view the allocation as using a blockSize=1, we can expose this blockSize for user to define e.g. up to 8/16, which would greatly reduce the lease collision?

ivelichkovich:
> NodeSlice CR is user to define and change...so hard to match the different runtime need of each node? not before the user is AI I guess. If we can view the current allocation as using a blockSize=1, we can expose this blockSize for user to define e.g. up to 8/16, which would greatly reduce the lease collision? and we will have node slice size based on need.

I'm not sure I fully understand. In this current version the user can define whatever slice size they need.

jingczhang:
Hi @ivelichkovich, sorry for not making my point clear. I meant to suggest not to limit a whereabouts node agent to only one network slice.  Here are more details for your review: (1) Limiting a node agent to one network slice effectively remove the need for "lease lock", since the locking will always be successful (2) We can use the existing "lease lock" workflow for the node agent to require access to other network slice (not full) when its primary slice is full.

ivelichkovich:
> Hi @ivelichkovich, sorry for not making my point clear. I meant to suggest not to limit a whereabouts node agent to only one network slice. Here are more details for your review: (1) Limiting a node agent to one network slice effectively remove the need for "lease lock", since the locking will always be successful (2) We can use the existing "lease lock" workflow for the node agent to require access to other network slice (not full) when its primary slice is full. (3) When a new node added, it can take a free network slice (not assigned to any node yet).

we discussed this in maintainers meeting, lease is still needed because you can run multiple network-attachment-definitions for same node as well as each node can allocate multiple IPs at same time and each launches a new whereabouts process

ivelichkovich:
note to self: clean imports

e2e/client/ippool.go
```diff
@@ -31,8 +32,38 @@ func isIPPoolAllocationsEmpty(ctx context.Context, k8sIPAM *kubeClient.Kubernete
...
+}
+
 // WaitForZeroIPPoolAllocations polls up to timeout seconds for IP pool allocations to be gone from the Kubernetes cluster.
 // Returns an error if any IP pool allocations remain after time limit, or if GETing IP pools causes an error.
 func WaitForZeroIPPoolAllocations(ctx context.Context, k8sIPAM *kubeClient.KubernetesIPAM, ipPoolCIDR string, timeout time.Duration) error {
 	return wait.PollUntilContextTimeout(ctx, time.Second, timeout, true, isIPPoolAllocationsEmpty(ctx, k8sIPAM, ipPoolCIDR))
 }
+
+// WaitForZeroIPPoolAllocationsAcrossNodeSlices polls up to timeout seconds for IP pool allocations to be gone from the Kubernetes cluster.
+// Returns an error if any IP pool allocations remain after time limit, or if GETing IP pools causes an error.
```
Eranbdr, line 66: Is there an option to get an explicit empty result from all the node slices, without waiting for the timeout?

e2e/e2e_node_slice_test.go
```diff
@@ -0,0 +1,353 @@
...
+			AfterEach(func() {
+				By("removing replicas and expecting 0 IP pool allocations")
+				Expect(
+					util.CheckZeroIPPoolAllocationsAndReplicas(
+						context.TODO(), clientInfo, k8sIPAM, rsName, testNamespace, ipv4TestRange, testNetworkName)).To(Succeed())
+
+				By("deleting replicaset with whereabouts net-attach-def")
+				Expect(clientInfo.DeleteReplicaSet(replicaSet)).To(Succeed())
+			})
+
```
Eranbdr, line 156: Can these tests be run from multi-threaded in parallel, to verify simultaneous requests from several client?

ivelichkovich: We could do something like that but what exactly are we trying to test with that? Whereabouts works on pod create so these tests do launch many concurrent pods.

Eranbdr: If I understand correctly the pod creation here is serial, so the IP requests toward whereabouts may be not fully concurrent, and so don't test simultaneous requests (at least if the test is blocking on pod creation)

ivelichkovich: So it sets the replicaset to `testConfig.MaxReplicas(allPods.Items)` that will result in many pods launching in parallel. Depending on how many nodes are in the test cluster this should also lead to multiple pods per node so this would get exercised with multiple pods to a lease/node pool.

ivelichkovich:
Might be worth marking this feature as experimental in the docs until we've built out more of the phases from the proposal and had more baketime/testing time

dougbtv (approved):
Igor -- this is awesome. I just ran through some functional testing, and it's looking great for the cases I tried, which didn't try to push the boundaries of the limitations we noted.

I'm all for moving forward with a merge. Especially since we've decided on a phased approach, if there's tailoring that's necessary, we can follow on with it, also because of the approach you took (and thank you for it), I don't think there's a strong risk to other functionality.

Agreed on marking it experimental in the docs... here's a quick attempt at an addition to the README. Feel free to incorporate it, or we can follow on with something:

```
## Fast IPAM by Using Preallocated Node Slices [Experimental]

**Enhance IPAM performance in large-scale Kubernetes environments by reducing IP allocation contention through node-based IP slicing.**

### Fast IPAM Configuration

metadata:
  name: whereabouts-fast-ipam
spec:
  config: '{
      "cniVersion": "0.3.0",
      "name": "whereaboutsexample",
      "type": "macvlan",
      "master": "eth0",
      "mode": "bridge",
      "ipam": {
        "type": "whereabouts",
        "range": "192.168.2.0/24",
        "fast_ipam": true,
        "node_slice size": "/22"
      }
    }'

This setup enables the fast IPAM feature to optimize IP allocation for nodes, improving network performance in clusters with high pod density.
```

mlguerrero12:
Hi @ivelichkovich, I'm about to start reviewing this PR but I wanted to understand the design first. From the proposal, it is not clear to me how the range is divided. 

Could you please elaborate how a range set in the IPAM config is divided between the nodes, assuming the `node_slice_size` represents a division larger than the current number of nodes? What would be the range in the `NodeSlicePool` for each node? i.e. range 192.168.1.0/24, node_slice_size /26, 2 nodes

What happens if the number of nodes increases? i.e. nodes increases to 6

What does the new controller do when a node is unreachable?

Thanks.

ivelichkovich:
> Hi @ivelichkovich, I'm about to start reviewing this PR but I wanted to understand the design first. From the proposal, it is not clear to me how the range is divided.
> 
> Could you please elaborate how a range set in the IPAM config is divided between the nodes, assuming the `node_slice_size` represents a division larger than the current number of nodes? What would be the range in the `NodeSlicePool` for each node? i.e. range 192.168.1.0/24, node_slice_size /26, 2 nodes
> 
> What happens if the number of nodes increases? i.e. nodes increases to 6
> 
> What does the new controller do when a node is unreachable?
> 
> Thanks.

Hey so this requires running a controller in the cluster, that controller is responsible for going for creating and managing the NodeSlicePools (resource representing node allocations). When nodes are added it assigns the nodes to a open "slice". If there's too many nodes it just skips them but it could fire an event or something like that. If a node is not reachable I don't think it'll be removed but when the node itself is actually deleted from the cluster it's "slice" will open up again.

pkg/types/types.go
```diff
@@ -81,6 +83,8 @@ func (ic *IPAMConfig) UnmarshalJSON(data []byte) error {
 		Datastore                string               `json:"datastore"`
 		Addresses                []Address            `json:"addresses,omitempty"`
 		IPRanges                 []RangeConfiguration `json:"ipRanges"`
+		NodeSliceSize            string               `json:"node_slice_size"`
+		Namespace                string               `json:"namespace"` //TODO: best way to get namespace of the NAD?
```
ivelichkovich, line 87: this is the biggest remaining issue in the PR I think, it's needed to know the namespace of the NAD which is the same namespace as the nodeslices so its used to get the nodeslices. 

Not sure if there's some easy way to discover this value, we could also make nodeslicepools be cluster scoped and not need to worry about it wouldn't be consistent with the rest of the CRDs.

ivelichkovich: It's also an API change so this one is probably worth figuring out even before merging as an experimental feature

ivelichkovich: okay fixed this to use WHEREABOUTS_NAMESPACE the same way it does for IPPools, there's some implications there if there's like multiple sets of duplicate NADs in different namespaces, maybe these resources should be cluster scoped? anyway for this PR this should fix the namespace thing to follow current patterns

dougbtv:
Appreciate all the hard work on this -- huge benefit to the whereabouts community, hugely appreciated.

PR #460, opened 2024-04-17
Proposal fast ipam

ivelichkovich:

proposal for fast IPAM through node slicing. Code PR: https://github.com/k8snetworkplumbingwg/whereabouts/pull/458  

Fixes #


doc/proposals/fast_ipam_by_node.md
```diff
@@ -0,0 +1,271 @@
...
+      "cniVersion": "0.3.0",
+      "name": "whereaboutsexample",
+      "type": "macvlan",
+      "master": "eth0",
+      "mode": "bridge",
+      "ipam": {
+        "type": "whereabouts",
+        "range": "192.168.2.225/28"
++       "fast_ipam": true,
++       "node_slice_size": "/22"
```
maiqueb, line 95: shouldn't the ranges be the other way around ? IIUC, you have a /22 range per node, but the range overall is /28 ...

I would envision something like a /16 range spread across nodes via /24 subnets - i.e. `"range": "192.168.0.0/16"` and `"node_slice_size": "/24"` .

Point being, the mask size on the range must be bigger than the node slice size.

Am I missing something ?

doc/proposals/fast_ipam_by_node.md
```diff
@@ -0,0 +1,271 @@
...
+#### whereabouts/pkg/types/types.go
+
+```diff
+ type IPAMConfig struct {
+      Name                string
+      Type                string               `json:"type"`
+      Routes              []*cnitypes.Route    `json:"routes"`
+      Datastore           string               `json:"datastore"`
+      IPRanges            []RangeConfiguration `json:"ipRanges"`
++     FastIpam            bool                 `json:"fast_ipam"`
```
maiqueb, line 136: is the boolean really required ? 

Can't we simply just infer if fast IPAM is used if the node slice size is provisioned ?

maiqueb (commented):
some initial questions that popped up.

doc/proposals/fast_ipam_by_node.md
ivelichkovich: Yeah sure we could do that

doc/proposals/fast_ipam_by_node.md
ivelichkovich: bad example for copy pasta another proposal, yes this current configuration would not work because node_slice_size is too big, I'll update with both the comments

ivelichkovich: fixed

doc/proposals/fast_ipam_by_node.md
ivelichkovich: updated per this feedback, ty I'll add this to the prototype today

doc/proposals/fast_ipam_by_node.md
```diff
@@ -0,0 +1,268 @@
...
++			return fmt.Sprintf("%s-%s", poolIdentifier.NetworkName, normalizeRange(poolIdentifier.IpRange))
++		}
+	}
+}
+
+```
+
+Additional changes will be required within whereabouts to use the NodeSlice to find the `Range` that the node its running
+on is assigned to. From here it can use the range on the IPPool with current code. Another change to set the IPPoolName and 
+Lease name as described above will be required. Finally, a new controller will be introduced to assign nodes to NodeSlices.
```
dougbtv, line 195: I'm definitely a-ok with the new controller. I think it's the right way forward.

doc/proposals/fast_ipam_by_node.md
```diff
@@ -0,0 +1,268 @@
...
+### Goals
+
+- Support existing whereabouts functionality without breaking changes
+- Introduce a new mode that can be configured on NAD to use IPAM by node slices
+- Support multiple NADs on same network range
+
+### Non-Goals
+
+- Migrate all users to this new mode
+
```
dougbtv, line 35: Potentially we can add here "integrate reconciliation controller and the new controller" as a non-goal -- that can be left as a follow on if it adds too much overhead. I'm ok with starting with a brand new controller for this process.

doc/proposals/fast_ipam_by_node.md
```diff
@@ -0,0 +1,268 @@
...
+### Backward Compatibility
+
+This feature will only change the behavior of whereabouts if the enabled flag is on. Otherwise 
+whereabouts will work the same for any IPAM config without the `node_slice_size` defined.
+
+## Alternative Design
+
+Another design is that the whereabouts daemonset that runs the install-cni script could be used to have a startup and 
+shutdown hook which would handle the assignment of nodes to a node slice. This would require locking for the `NodeSlicePools`
+on Node join. The reason to use the controller over this design is because the reconciliation pattern reduces the likelyhood for bugs (like leaked IPs)
```
dougbtv, line 255: +1

dougbtv:
One thing I'm kind of thinking about is that the reconciler has this "audit style reconciliation" that runs on a timed fashion. I'm wondering if we might introduce resource contention with it with the fast style allocation since it won't be using the leases. Just thinking out loud

doc/proposals/fast_ipam_by_node.md
ivelichkovich: yes, I was actually thinking about adding this as a non-goal but we may want to do it in the future (with changes to the existing controller) so I left it as is for now but I can add it

ivelichkovich:
> One thing I'm kind of thinking about is that the reconciler has this "audit style reconciliation" that runs on a timed fashion. I'm wondering if we might introduce resource contention with it with the fast style allocation since it won't be using the leases. Just thinking out loud

Are you referring to the gocron running within the existing reconciler controller? That one doesn't use leases at all as far as I can tell, it calls allocation.IterateForDeallocation directly.

dougbtv:
> > One thing I'm kind of thinking about is that the reconciler has this "audit style reconciliation" that runs on a timed fashion. I'm wondering if we might introduce resource contention with it with the fast style allocation since it won't be using the leases. Just thinking out loud
> 
> Are you referring to the gocron running within the existing reconciler controller? That one doesn't use leases at all as far as I can tell, it calls allocation.IterateForDeallocation directly.

So! That being the case... we don't need to worry about it. I think that might just be an existing issue, so, I will look into that, thanks for taking a peek.

doc/proposals/fast_ipam_by_node.md
```diff
@@ -0,0 +1,268 @@
...
+
+- Support existing whereabouts functionality without breaking changes
+- Introduce a new mode that can be configured on NAD to use IPAM by node slices
+- Support multiple NADs on same network range
+
+### Non-Goals
+
+- Migrate all users to this new mode
+
+<hr>
```
dougbtv, line 36: I think we're good to go on this design! Let's just add a section for next steps, may I suggest

```
## Phases

Initial phase:
[x] Fast ranges base functionality

Feature parity phase:
[ ] Range start, range-end function
[ ] Live changes to range (from regular to fast)
[ ] Multiple ranges

Optimization phase:
[ ] Dynamic range slice allocation
```

Definitely feel free to add / remove / edit these, just to get a placeholdre, thanks Igor!

ivelichkovich: awesome thanks for the help! Let me just add that nice diagram as well from the maintainers call and then we should be good to go!

ivelichkovich: woops I thought you committed this in already, just pushed the suggestions

PR #467, opened 2024-05-03
Use IP to identify orphaned allocation to be deleted 

mlguerrero12:
- Del command should not fail when allocation is not found
- The IP uniquely identifies the correct orphaned allocation that needs to be deleted during IP reconcilation.

Fixes https://github.com/k8snetworkplumbingwg/whereabouts/issues/176

### DIFF
--- a/cmd/whereabouts.go
+++ b/cmd/whereabouts.go
@@ -19,14 +19,14 @@ import (
 
 func cmdAddFunc(args *skel.CmdArgs) error {
 	ipamConf, confVersion, err := config.LoadIPAMConfig(args.StdinData, args.Args)
-    if err != nil {
+	if err != nil {
 		logging.Errorf("IPAM configuration load failed: %s", err)
 		return err
 	}
 	logging.Debugf("ADD - IPAM configuration successfully read: %+v", *ipamConf)
 	ipam, err := kubernetes.NewKubernetesIPAM(args.ContainerID, *ipamConf)
 	if err != nil {
-			return logging.Errorf("failed to create Kubernetes IPAM manager: %v", err)
+		return logging.Errorf("failed to create Kubernetes IPAM manager: %v", err)
 	}
 	defer func() { safeCloseKubernetesBackendConnection(ipam) }()
 	return cmdAdd(args, ipam, confVersion)
@@ -48,15 +48,14 @@ func cmdDelFunc(args *skel.CmdArgs) error {
 	return cmdDel(args, ipam)
 }
 
-
 func main() {
 	skel.PluginMainFuncs(skel.CNIFuncs{
-		Add:    cmdAddFunc,
-		Check:  cmdCheck,
-		Del:    cmdDelFunc,
-	}, 
-	cniversion.All, 
-	fmt.Sprintf("whereabouts %s", version.GetFullVersionWithRuntimeInfo()))
+		Add:   cmdAddFunc,
+		Check: cmdCheck,
+		Del:   cmdDelFunc,
+	},
+		cniversion.All,
+		fmt.Sprintf("whereabouts %s", version.GetFullVersionWithRuntimeInfo()))
 }
 
 func safeCloseKubernetesBackendConnection(ipam *kubernetes.KubernetesIPAM) {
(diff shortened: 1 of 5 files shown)


pkg/allocate/allocate.go
```diff
@@ -36,46 +36,29 @@ func AssignIP(ipamConf types.RangeConfiguration, reservelist []types.IPReservati
...
-	foundidx := -1
 	for idx, v := range reservelist {
 		if v.ContainerID == id {
-			foundidx = idx
-			break
+			return idx
 		}
 	}
-	return foundidx
+	return -1
```
dougbtv, line 61: good eye on the efficiency :+1:

pkg/storage/kubernetes/ipam.go
```diff
@@ -542,10 +542,10 @@ func IPManagementKubernetesUpdate(ctx context.Context, mode int, ipam *Kubernete
 				}
 
 			case whereaboutstypes.Deallocate:
-				updatedreservelist, ipforoverlappingrangeupdate, err = allocate.DeallocateIP(reservelist, containerID)
-				if err != nil {
-					logging.Errorf("Error deallocating IP: %v", err)
-					return newips, err
+				updatedreservelist, ipforoverlappingrangeupdate = allocate.DeallocateIP(reservelist, containerID)
+				if updatedreservelist == nil {
+					// Do not fail if allocation is not present.
```
dougbtv, line 547: Cool, I'm following. I like that you logged the condition in `allocate.DeallocateIP()` -- thanks -- and can silently pass here otherwise. Nice one.

dougbtv (approved):
LGTM! Thank you Marcelo

cmd/whereabouts.go
```diff
@@ -110,10 +109,7 @@ func cmdDel(args *skel.CmdArgs, client *kubernetes.KubernetesIPAM) error {
 	ctx, cancel := context.WithTimeout(context.Background(), types.DelTimeLimit)
 	defer cancel()
 
-	_, err := kubernetes.IPManagement(ctx, types.Deallocate, client.Config, client)
-	if err != nil {
-		logging.Verbosef("WARNING: Problem deallocating IP: %s", err)
-	}
+	_, _ = kubernetes.IPManagement(ctx, types.Deallocate, client.Config, client)
```
pliurh, line 112: If we don't use any of the return values. Shall we remove the assignment to the blank identifier?

pliurh:
@mlguerrero12 It looks good to me. Just on a side note. Could you also change the function name of `isPodAlive`? It's misleading. It should be something like`isOrphanedIP`.

pkg/storage/kubernetes/ipam.go
maiqueb: that's exactly what I don't like about it lol. 

I'm ok with it though, since we do not unit test that function.

maiqueb (commented):
I just have a couple of opinionated nits. 

You can merge it without my approval, I don't have better suggestions.

cmd/whereabouts.go
mlguerrero12: This is the proper way to ignore returned values. I prefer to have it like this.

pkg/allocate/allocate.go
```diff
@@ -36,17 +36,20 @@ func AssignIP(ipamConf types.RangeConfiguration, reservelist []types.IPReservati
...
 }
 
-// DeallocateIP assigns an IP using a range and a reserve list.
-func DeallocateIP(reservelist []types.IPReservation, containerID string) ([]types.IPReservation, net.IP, error) {
-
-	updatedreservelist, hadip, err := IterateForDeallocation(reservelist, containerID, getMatchingIPReservationIndex)
-	if err != nil {
-		return nil, nil, err
+// DeallocateIP removes allocation from reserve list.
+func DeallocateIP(reservelist []types.IPReservation, containerID string) ([]types.IPReservation, net.IP) {
```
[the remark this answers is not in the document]
mlguerrero12: I see your point but I still don't think an error should be returned. However, I will change the logic a bit and return nil for IP when the allocation is not present. It makes more sense in my opinion. When not found, return the reverlist untouched and nil for IP.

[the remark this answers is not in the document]
maiqueb: it's fine the way you currently have. Don't worry. If we had some unit tests at least to define the desired behavior, it could be arguable. Since we don't, let's just use the simplest thing. 

I'm fine w/ what you have. Really.

mlguerrero12:
@maiqueb, @pliurh, thanks for you comments. I addressed them. Please have a look again.

dougbtv (approved):
LGTM! Thank you

PR #490, opened 2024-07-31
Simple scale script

bpickard22:
Adds a script that can be used with the the local kind cluster to test scale impact of local changes


dougbtv (changes requested):
let's add a note: https://github.com/k8snetworkplumbingwg/whereabouts/blob/master/doc/developer_notes.md

dougbtv (approved):
lgtm! thanks Ben

PR #495, opened 2024-08-20
Kubeconfig loop

bpickard22:
Creates functions to generate the kubeconfig and whereabouts conf use these functions to check if the serviveAccount changes to regen the kubeconfig if it expires

Change WHEREABOUTS_FLATFILE to WHEREABOUTS_CONF_FILE for clarity


bpickard22:
/hold
Script is currently breaks kind cluster startup
Also still need to add testing so holding merge until im done with that

PR #502, opened 2024-09-02
Add missing crds and options in the chart

thomasferrandiz:
This PR adds some updates from the whereabouts template that were missing in the Helm chart.


mlguerrero12:
I think we should modify hack/generate-code.sh to not only places the crds in doc/crds but also in the charts folder. This improves maintainability

mlguerrero12 (approved):

dougbtv (approved):

Issue #498, opened 2024-08-27
[BUG] Node Slice Fast IPAM Uses Incorrect Range Field When Specs Change


ivelichkovich:
**Describe the bug**

https://github.com/k8snetworkplumbingwg/whereabouts/blob/master/pkg/node-controller/controller.go#L436 should use `ipamConf.IPRanges[0].Range` instead of `ipamConf.Range` like the if block above it that checks for the change to spec

**Expected behavior**

Whereabouts node controller should handle spec changes

**To Reproduce**
1.  Create a NAD
2. Change node_slice_size or ip_range
3.

**Environment**:
- Whereabouts version : latest main
- Kubernetes version (use `kubectl version`): N/A
- Network-attachment-definition: N/A
- Whereabouts configuration (on the host): N/A
- OS (e.g. from /etc/os-release): N/A
- Kernel (e.g. `uname -a`): N/A
- Others: N/A

**Additional info / context**
Add any other information / context about the problem here.

ivelichkovich:
Thanks @xagent003 for finding this, please help raise MR to fix 👍

PR #503, opened 2024-09-13
Fixes for node slice IPAM

xagent003:
This fixes some issues seen with the new node slice IPAM feature.

1. Disable some controllers that are not needed. Don't need informer on nodeSlicePool since that is an internal CR we manage. Don't need to reconcile for NADs when the resource version does not change. For nodes, only listen for node add and delete. Nodes are updated frequently for Status changes and conditions that we don't need - we only want to allocate a new slice pool when a node is created, and remove its allocation when node is deleted

2. We have multiple NADs(which map to multiple NICs), but with the same CIDR and network_name(because it is really one L2 network). With node slice pool feature enabled and a Pod requesting multiple networks, the same podRef and same containerID will be present multiple times in each IPpool, for each ifName(corresponding to each NAD). We also need to match by ifName so it deletes the correct entry, rather than first one

3. When node slice size or cidr is reconfigured, it was passing the wrong range - ipam.Range, rather than ipamConf.IPRanges[0].Range. ipam.Range is cleared and set to ipamConf.IPRanges[0].Range so it was error'ing out with empty CIDR

4. nodeSlice.Spec was not being written/saved when the node slice size/cidr was reconfigured

5. the Subnet mask/cidr used was incorrect. We should use the CIDR from the NAD, rather than the range for the node. The NAD's range really defines the cluster-wide subnet, whereas the nodeSLicePool's IPRange is just used for grouping IP allocations. Without this fix, each node was on a different subnet, resulting in a IP lookup via default route(primary CNI), rather than going over the NAD/Multus network. For example, suppose our NAD range is 10.0.0.0/8. The node slice size is a /24. If we use the range from the NodeSLicePool, Pods on a node are on a different /24, rather than all being on the same /8.

### DIFF
--- a/pkg/allocate/allocate.go
+++ b/pkg/allocate/allocate.go
@@ -50,8 +50,8 @@ func AssignIP(ipamConf types.RangeConfiguration, reservelist []types.IPReservati
 }
 
 // DeallocateIP removes allocation from reserve list. Returns the updated reserve list and the deallocated IP.
-func DeallocateIP(reservelist []types.IPReservation, containerID string) ([]types.IPReservation, net.IP) {
-	index := getMatchingIPReservationIndex(reservelist, containerID)
+func DeallocateIP(reservelist []types.IPReservation, containerID, ifName string) ([]types.IPReservation, net.IP) {
+	index := getMatchingIPReservationIndex(reservelist, containerID, ifName)
 	if index < 0 {
 		// Allocation not found. Return the original reserve list and nil IP.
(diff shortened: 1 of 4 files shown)


ivelichkovich:
/lgtm 

Thank you!

ivelichkovich:
fixes: https://github.com/k8snetworkplumbingwg/whereabouts/issues/498

dougbtv (approved):
Thank you Arjun!!!

PR #505, opened 2024-09-18
Reduce Sleep

bpickard22:
during the reconcilation loop we check to see if we have any orphaned Ips, and during this we currently have a 1.5 second sleep. I ran some tests and found a decent perf improvement by dropping this by half, but removing it completely caused issues


bpickard22:
PTAL @dougbtv

mlguerrero12 (approved):

dougbtv (approved):

PR #512, opened 2024-10-15
Bump Go version 1.23

mlguerrero12:
- Align with version used in Dockerfile
-  Workflows use go version set in go.mod
- Fix overlappingiprange test

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,19 +4,18 @@ jobs:
   build:
     strategy:
       matrix:
-        go-version: [1.21.x]
         #goarch: [386, amd64, arm, ppc64le, arm64]
         goarch: [amd64, arm64]
         os: [ubuntu-latest] #, macos-latest, windows-latest]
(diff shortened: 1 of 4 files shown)


thomasferrandiz (approved):

dougbtv (approved):
lgtm. I see a failing e2e so I tried restarting it in case it's a flake

mlguerrero12:
It seems to be a consistent issue. I will investigate

mlguerrero12:
@dougbtv, I fixed the overlappingiprange test. It was constantly failing on my local setup as well. This test has a for loop that sets the var `enableOverlappingRanges` to `true` or `false`. Based on this, it checks if the ips of the pods are equal or different. The issue is that it never created the new NAD using this var. It was always false. So, it ran the same scenario but with different checks at the end. Did not investigate why this test has been passing. It's a bit strange, but it is good now.

Please have a look again.

dougbtv (approved):

PR #516, opened 2024-11-25
Skip Static test

bpickard22:
In order to run upstream tests in a container with make test we need to be able to skip the static check

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ COMPUTE_NODES ?= 2
 
 OCI_BIN ?= docker
 
+
 build:
 	hack/build-go.sh
 
@@ -23,7 +24,10 @@ install-tools:
 	hack/install-kubebuilder-tools.sh
 
 test: build install-tools
-	hack/test-go.sh
+	hack/test-go.sh 
+
+test-skip-static: build
+	hack/test-go.sh --skip-static-check 
 
 kind:
 	hack/e2e-setup-kind-cluster.sh -n $(COMPUTE_NODES)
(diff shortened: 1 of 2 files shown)


manuelbuil:
What wrong with staticcheck when run in a container?

Makefile
```diff
@@ -23,7 +25,10 @@ install-tools:
 	hack/install-kubebuilder-tools.sh
 
 test: build install-tools
-	hack/test-go.sh
+	hack/test-go.sh --skip-static-check
```
dougbtv, line 28: wait, isn't this going to skip the static check but, we actually want to run it here, right?

bpickard22: no since the default is false

bpickard22: I tested this on my local the logic behaves as expected, if you dont pass in skip-static-check arg it will error out, the way its written currently its not an optional var

bpickard22:
> What wrong with staticcheck when run in a container?

If we want to call make test in a container running in a cluster we cannot assume that static-check will be installed, and you cant run the `go install` command in -vendor mode 

This is currently an issue that we are hitting in our (RedHat) downstream testing

dougbtv (approved):
thanks! looking good

dougbtv:
> What wrong with staticcheck when run in a container?

I think Ben's description is incomplete, I think he needs a make target where he can skip this because he's trying to test it a disconnected environment where he can't pull that gomodule, so he needs a target where he can skip it

dougbtv:
Thank you!

Issue #483, opened 2024-06-20
[BUG] Scale down of deployment left pod references undeleted


smoshiur1237:
**Describe the bug**
We are having a situation in our test cluster where Pod references left undeleted during scale down operation of a deployment. It get increased if we do multiple scale down/up operation. After scaling down of deployment from 200 to 1, most of the pods stays in `Terninating` state and needs long time to get deleted from the cluster. After the deletion some of the pod references are still visible and may increase after multiple scale up/down. Here is listed some queries during the process:

Scale down form 200 to 1 and when most of the pods are in `Terminating` state
```
date && kubectl get ippools.whereabouts.cni.cncf.io -n kube-system  2.2.2.0-24 -o yaml | grep -c podref 
Thu Jun 20 06:56:08 UTC 2024
161
date && kubectl get ippools.whereabouts.cni.cncf.io -n kube-system  3.3.3.0-24 -o yaml | grep -c podref
Thu Jun 20 06:53:34 UTC 2024
121

date && kubectl get ippools.whereabouts.cni.cncf.io -n kube-system  4.4.4.0-24 -o yaml | grep -c podref 
Thu Jun 20 06:53:55 UTC 2024
66

date && kubectl get ippools.whereabouts.cni.cncf.io -n kube-system  5.5.5.0-24 -o yaml | grep -c podref
Thu Jun 20 06:54:36 UTC 2024
11

kubectl get ippools.whereabouts.cni.cncf.io -n kube-system  6.6.6.0-24 -o yaml | grep -c podref 
Thu Jun 20 06:54:56 UTC 2024
1

kubectl get ippools.whereabouts.cni.cncf.io -n kube-system  7.7.7.0-24 -o yaml | grep -c podref 
Thu Jun 20 06:55:15 UTC 2024
1

date && kubectl get ippools.whereabouts.cni.cncf.io -n kube-system  8.8.8.0-24 -o yaml | grep -c podref 
Thu Jun 20 06:55:35 UTC 2024
1

kubectl get ippools.whereabouts.cni.cncf.io -n kube-system  9.9.9.0-24 -o yaml | grep -c podref 
Thu Jun 20 06:55:55 UTC 2024
1
```
When scale down from 200 to 1 is completed and all replicas are deleted
```
date && kubectl get ippools.whereabouts.cni.cncf.io -n kube-system  2.2.2.0-24 -o yaml | grep -c podref
Thu Jun 20 08:30:53 UTC 2024
2
[...]

PR #480, opened 2024-06-14
Align api calls timeouts cronjob ip reconciler

mlguerrero12:
Parent timeout context of 30s was removed. All listing operations used by the cronjob reconciler has 30s as timeout.

Fixes https://github.com/k8snetworkplumbingwg/whereabouts/issues/389

PR #556, opened 2025-02-06
🐛 Fixes leftover podref issue

adilGhaffarDev:
Leader election fails when `IPManagement` is called from `garbageCollectPodIPs` in `pod.go`, this PR fixes the `KubernetesIPAM` and `Client` creation in `pod.go`, which resolves the leader election issue.

Fixes https://github.com/k8snetworkplumbingwg/whereabouts/issues/483

### DIFF
--- a/pkg/controlloop/pod.go
+++ b/pkg/controlloop/pod.go
@@ -224,17 +224,16 @@ func (pc *PodController) garbageCollectPodIPs(pod *v1.Pod) error {
 				if allocation.PodRef == podID(podNamespace, podName) {
 					logging.Verbosef("stale allocation to cleanup: %+v", allocation)
 
-					client := *wbclient.NewKubernetesClient(nil, pc.k8sClient)
-					wbClient := &wbclient.KubernetesIPAM{
-						Client: client,
-						Config: *ipamConfig,
+					client := *wbclient.NewKubernetesClient(pc.wbClient, pc.k8sClient)
+					k8sIPAM := &wbclient.KubernetesIPAM{
+						Config:      *ipamConfig,
+						ContainerID: allocation.ContainerID,
+						IfName:      allocation.IfName,
+						Namespace:   pool.Namespace,
+						Client:      client,
 					}
 
-					if err != nil {
-						logging.Debugf("error while generating the IPAM client: %v", err)
-						continue
-					}
-					if _, err := pc.cleanupFunc(context.TODO(), types.Deallocate, *ipamConfig, wbClient); err != nil {
+					if _, err := pc.cleanupFunc(context.TODO(), types.Deallocate, *ipamConfig, k8sIPAM); err != nil {
 						logging.Errorf("failed to cleanup allocation: %v", err)
 					}
 					if err := pc.addressGarbageCollected(pod, nad.GetName(), pool.Spec.Range, allocationIndex); err != nil {
@@ -244,7 +243,6 @@ func (pc *PodController) garbageCollectPodIPs(pod *v1.Pod) error {
 			}
 		}
 	}
-
 	return nil
 }
 
--- a/pkg/storage/kubernetes/ipam.go
+++ b/pkg/storage/kubernetes/ipam.go
@@ -36,17 +36,17 @@ const UnnamedNetwork string = ""
 type KubernetesIPAM struct {
 	Client
 	Config      whereaboutstypes.IPAMConfig
-	namespace   string
-	containerID string
+	Namespace   string
+	ContainerID string
 	IfName      string
 }
 
 func newKubernetesIPAM(containerID, ifName string, ipamConf whereaboutstypes.IPAMConfig, namespace string, kubernetesClient Client) *KubernetesIPAM {
 	return &KubernetesIPAM{
 		Config:      ipamConf,
-		containerID: containerID,
+		ContainerID: containerID,
 		IfName:      ifName,
-		namespace:   namespace,
+		Namespace:   namespace,
 		Client:      kubernetesClient,
 	}
 }
@@ -76,7 +76,7 @@ func NewKubernetesIPAMWithNamespace(containerID, ifName string, ipamConf whereab
 	if err != nil {
 		return nil, err
 	}
-	k8sIPAM.namespace = namespace
+	k8sIPAM.Namespace = namespace
 	return k8sIPAM, nil
 }
 
@@ -137,14 +137,14 @@ func (i *KubernetesIPAM) getPool(ctx context.Context, name string, iprange strin
 	ctxWithTimeout, cancel := context.WithTimeout(ctx, storage.RequestTimeout)
 	defer cancel()
 
-	pool, err := i.client.WhereaboutsV1alpha1().IPPools(i.namespace).Get(ctxWithTimeout, name, metav1.GetOptions{})
+	pool, err := i.client.WhereaboutsV1alpha1().IPPools(i.Namespace).Get(ctxWithTimeout, name, metav1.GetOptions{})
 	if err != nil && errors.IsNotFound(err) {
 		// pool does not exist, create it
 		newPool := &whereaboutsv1alpha1.IPPool{}
 		newPool.ObjectMeta.Name = name
 		newPool.Spec.Range = iprange
 		newPool.Spec.Allocations = make(map[string]whereaboutsv1alpha1.IPAllocation)
-		_, err = i.client.WhereaboutsV1alpha1().IPPools(i.namespace).Create(ctxWithTimeout, newPool, metav1.CreateOptions{})
+		_, err = i.client.WhereaboutsV1alpha1().IPPools(i.Namespace).Create(ctxWithTimeout, newPool, metav1.CreateOptions{})
 		if err != nil && errors.IsAlreadyExists(err) {
 			// the pool was just created -- allow retry
 			return nil, &temporaryError{err}
@@ -162,7 +162,7 @@ func (i *KubernetesIPAM) getPool(ctx context.Context, name string, iprange strin
 
 // Status tests connectivity to the kubernetes backend
 func (i *KubernetesIPAM) Status(ctx context.Context) error {
-	_, err := i.client.WhereaboutsV1alpha1().IPPools(i.namespace).List(ctx, metav1.ListOptions{})
+	_, err := i.client.WhereaboutsV1alpha1().IPPools(i.Namespace).List(ctx, metav1.ListOptions{})
 	return err
 }
 
@@ -276,7 +276,7 @@ type KubernetesOverlappingRangeStore struct {
 
 // GetOverlappingRangeStore returns a clusterstore interface
 func (i *KubernetesIPAM) GetOverlappingRangeStore() (storage.OverlappingRangeStore, error) {
-	return &KubernetesOverlappingRangeStore{i.client, i.namespace}, nil
+	return &KubernetesOverlappingRangeStore{i.client, i.Namespace}, nil
 }
 
 // IsAllocatedInOverlappingRange checks for IP addresses to see if they're allocated cluster wide, for overlapping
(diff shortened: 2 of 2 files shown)


adilGhaffarDev:
@thomasferrandiz please check

adilGhaffarDev:
This patch works perfectly on top of version 0.8.0, but it still does not work on the main branch.

I would like to ask if it's possible to create a patch release (0.8.1) that includes this fix.

Additionally, I tried using Node Slice Fast IPAM since it addresses leader election issues, but instead of helping, it worsened the situation. I have opened an issue explaining the problem: [#558](https://github.com/k8snetworkplumbingwg/whereabouts/issues/558). Please take a look.

maiqueb:
> This patch works perfectly on top of version 0.8.0, but it still does not work on the main branch.

I think this is not how it should be. You need to make your fix to the main branch, and once it works, you back-port it to the release branch.

> 
> I would like to ask if it's possible to create a patch release (0.8.1) that includes this fix.
> 
> Additionally, I tried using Node Slice Fast IPAM since it addresses leader election issues, but instead of helping, it worsened the situation. I have opened an issue explaining the problem: [#558](https://github.com/k8snetworkplumbingwg/whereabouts/issues/558). Please take a look.

adilGhaffarDev:
> I think this is not how it should be. You need to make your fix to the main branch, and once it works, you back-port it to the release branch.

but there is no 0.8.0 branch and code changed quite a lot since 0.8.0 with addition of fast ipam node slice, so main will need a different fix. Is there a way to merge just this fix on 0.8.0? I know merging it on main is not going to help, I opened the PR just to share my patch.

/hold

maiqueb:
I think we should cut a release branch for 0.8 as you're pointing out.

@dougbtv could you assist in that? What do you think ?

adilGhaffarDev:
> I think we should cut a release branch for 0.8 as you're pointing out.

yes, that would be great.

pkg/controlloop/pod.go
```diff
@@ -224,27 +225,53 @@ func (pc *PodController) garbageCollectPodIPs(pod *v1.Pod) error {
...
-					if err != nil {
-						logging.Debugf("error while generating the IPAM client: %v", err)
-						continue
-					}
-					if _, err := pc.cleanupFunc(context.TODO(), types.Deallocate, *ipamConfig, wbClient); err != nil {
-						logging.Errorf("failed to cleanup allocation: %v", err)
-					}
-					if err := pc.addressGarbageCollected(pod, nad.GetName(), pool.Spec.Range, allocationIndex); err != nil {
-						logging.Errorf("failed to issue event for successful IP address cleanup: %v", err)
+					podExist ,_ := podExists(wbClient, podNamespace, podName)
```
mlguerrero12, line 235: If I'm not mistaken, the informer has only a handler for pod deletion. So, pods won´t exist. As far as I remember, the idea of this controller is to delete allocations that were not processed during the regular CNI DEL command. When a pod is force deleted or when there are multiple delete operations (giving up waiting for the leader election).

So, when this happens, you are always deleting directly without relaying on the IP management which may cause other operations in the pool to fail. A CNI ADD or DEL command that might be running in parallel. Those operations have retries but here I see you are only logging the error. So, your update can fail and no retries will happen. Leaving orphaned IPs.

Why do you believe that the issue is due to a deleted pod? Is it because the "IPAM client initialization error: no pod name" log? That shouldn't be a problem since the check in line 224 only succeeds if there is a valid name.

mlguerrero12:
> > I think this is not how it should be. You need to make your fix to the main branch, and once it works, you back-port it to the release branch.
> 
> but there is no 0.8.0 branch and code changed quite a lot since 0.8.0 with addition of fast ipam node slice, so main will need a different fix. Is there a way to merge just this fix on 0.8.0? I know merging it on main is not going to help, I opened the PR just to share my patch.
> 
> /hold

There shouldn't be major changes. The fast ipam node slice was meant to be a optional feature. Just for experimentation purposes. We should aim at fixing master primarily

adilGhaffarDev:
> There shouldn't be major changes. The fast ipam node slice was meant to be a optional feature. Just for experimentation purposes. We should aim at fixing master primarily

It looks like the pod controller is not running at all on `main branch`. I don’t see any logs from pkg/controlloop/pod.go, and I’m unsure why it isn’t running. Or maybe it's running but not called from anywhere.

pkg/controlloop/pod.go
mlguerrero12: Your comment about the deleted pods implies that we could also apply it to the CNI DEL command, but the idea of having the IP management logic is to avoid collisions. Say you delete 500 pods at once.

My point is that you are changing the whole logic here and not a particular part of it. This part is relaying on the IP management logic. Someone can force delete the 500 pods. All of them will be handled here. Have you tried this scenario? I also don't know how much we rely on this. When we delete 500 pods, how many allocations are handled here in this controller and not by the CNI DEL commands.

Something else seems to be going on. Let's understand first why we are having those log messages.

mlguerrero12:
That's a different issue. I'll have a look as well. But that shouldn't be a reason to create a fix for a tag and not in the master branch.

pkg/controlloop/pod.go
adilGhaffarDev: I have updated the patch, ensuring it does not alter any logic. It only fixes the `KubernetesIPAM` and `Client` creation in `pod.go`, which resolves the leader election issue. I have tested it on version `0.8.0` it solves the issue and I will test it on `main`, then report back.

dougbtv:
Really appreciate the contrib, and I'm definitely open to making a v0.8.1 and releasing it.

However, I'd really appreciate the eye towards getting it fixed against the current master branch. I will add this to the maintainer's agenda for next week (which I will miss, but can follow up subsequently)

adilGhaffarDev:
> However, I'd really appreciate the eye towards getting it fixed against the current master branch. I will add this to the maintainer's agenda for next week (which I will miss, but can follow up subsequently)

yes, I will try to fix it on the main too. 
Regarding this PR, it not only fixes the issue in 0.8.0, it's also fixing the following broken code in `pod.go`

https://github.com/k8snetworkplumbingwg/whereabouts/blob/d3633fa680bd4b8b5e8425b5a643fed1eec5f568/pkg/controlloop/pod.go#L226C1-L242C7

adilGhaffarDev:
@mlguerrero12 kindly check this when you get time.

mlguerrero12:
@adilGhaffarDev, I also see that the pod controller is not running on master. Will provide a fix soon

adilGhaffarDev:
> @adilGhaffarDev, I also see that the pod controller is not running on master. Will provide a fix soon

Thank you. This fix should still be needed. Can we merge this too once you have fixed the pod controller?

mlguerrero12:
yes, I will review and test these changes after we fix master

mlguerrero12:
@adilGhaffarDev, I still need to test some things from the current code, but it seems to me you are not fixing the issue with the leader election (occasional missing of pod names) but some missing data needed in `IPManagementKubernetesUpdate` to perform the deallocation. Am I seeing this correctly?

adilGhaffarDev:
> @adilGhaffarDev, I still need to test some things from the current code, but it seems to me you are not fixing the issue with the leader election (occasional missing of pod names) but some missing data needed in `IPManagementKubernetesUpdate` to perform the deallocation. Am I seeing this correctly?

When cleanupFunc is called from garbageCollectPodIPs for deallocation, it fails due to improperly initialized client and wbClient in pod.go. Specifically, it fails at leader election.

The old code initializes the `client` and `wbClient` incorrectly, leading to issues with leader election:
```
client := *wbclient.NewKubernetesClient(nil, pc.k8sClient)
wbClient := &wbclient.KubernetesIPAM{
				Client: client,
				Config: *ipamConfig,
			}
```

This PR ensures proper initialization so that leader election and deallocation work correctly:
```
client := *wbclient.NewKubernetesClient(pc.wbClient, pc.k8sClient)
k8sIPAM := wbclient.NewKubernetesIPAMWithClient(allocation.ContainerID, allocation.IfName, *ipamConfig, pool.Namespace, client)
```

With this fix, cleanupFunc receives correctly initialized parameters, allowing leader election to function properly and ensuring successful deallocation.

mlguerrero12:
I can also see that there are some stuff missing. So, `pc.cleanupFunc` is `IPManagement`. This ones calls `newLeaderElector`

```
	le, leader, deposed := newLeaderElector(client.clientSet, client.namespace, ipamConf.PodNamespace, ipamConf.PodName, ipamConf.LeaderLeaseDuration, ipamConf.LeaderRenewDeadline, ipamConf.LeaderRetryPeriod)

```

client.clientSet is the k8s client that is already defined but didn't realize until now that client.namespace is not set. So, yes, this one is missing for the leader election part.

After that, it calls `IPManagementKubernetesUpdate`. This one need besides the containterID for deallocation, the whereabouts client that is not set. These things are missing.

Your findings are correct. Just wanted to clarify the different sections.

Good catch @adilGhaffarDev

pkg/controlloop/pod.go
```diff
@@ -224,17 +224,10 @@ func (pc *PodController) garbageCollectPodIPs(pod *v1.Pod) error {
 				if allocation.PodRef == podID(podNamespace, podName) {
 					logging.Verbosef("stale allocation to cleanup: %+v", allocation)
 
-					client := *wbclient.NewKubernetesClient(nil, pc.k8sClient)
-					wbClient := &wbclient.KubernetesIPAM{
-						Client: client,
-						Config: *ipamConfig,
-					}
+					client := *wbclient.NewKubernetesClient(pc.wbClient, pc.k8sClient)
+					k8sIPAM := wbclient.NewKubernetesIPAMWithClient(allocation.ContainerID, allocation.IfName, *ipamConfig, pool.Namespace, client)
```
mlguerrero12, line 228: let's avoid having another abstraction here. Please add the missing info in the structure directly (Line 228). Let's keep the new name k8sIPAM or just ipam instead of wbClient which is confusing.

pkg/storage/kubernetes/ipam.go
```diff
@@ -70,6 +70,17 @@ func NewKubernetesIPAM(containerID, ifName string, ipamConf whereaboutstypes.IPA
 	return k8sIPAM, nil
 }
 
+// NewKubernetesIPAMWithClient returns a new KubernetesIPAM Client configured to a kubernetes CRD backend
+func NewKubernetesIPAMWithClient(containerID, ifName string, ipamConf whereaboutstypes.IPAMConfig, namespace string, kubernetesClient Client) *KubernetesIPAM {
```
mlguerrero12, line 74: no need to have this. We could declare it above as I commented

pkg/controlloop/pod.go
mlguerrero12, line 228: I'm doubting about the namespace being pool.Namespace.
To be consistent with the namespace obtained via the CNI command (from the loaded config - current context - ipam.go `NewKubernetesIPAM`), we could try to do the same in controlloop.go `newPodController`. After loading the config, you could get the namespace from the context and store this namespace as a new variable in the `PodController` structure. Then here you can do pc.Namespace for example.

adilGhaffarDev: Can I use `ipPoolsNamespace()` function? 
In `controlloop.go`, what would be the best way to load ipam config?  In the case of NewKubernetesIPAM, we get it from `whereabouts.go` `cmdAddFunc`. Not sure whats the best way to get it in `controlloop.go`

pkg/storage/kubernetes/ipam.go
adilGhaffarDev: done

pkg/controlloop/pod.go
mlguerrero12: Namespace from the pool is fine

adilGhaffarDev:
> @adilGhaffarDev, could you please rebase?

done

adilGhaffarDev:
after the merge of #577 I have tested this PR on main branch and the issue is fixed on main too. 
I am following the same steps explained here: https://github.com/k8snetworkplumbingwg/whereabouts/issues/483#issue-2364341172 , here is in more detail how to reproduce the issue: https://github.com/k8snetworkplumbingwg/whereabouts/pull/480#issuecomment-2178702545

PR #563, opened 2025-02-14
Host name determination from k8s api

bpickard22:
Host name in /etc/host can differ from the nodename in the k8s API, so we standardize the way we calculate the hostname -- always from the K8s API, and then write it to a file on the host so that the CNI plugin side can always read it.

Contributer: @DOUGBTV the goat

### DIFF
--- a/pkg/storage/kubernetes/ipam.go
+++ b/pkg/storage/kubernetes/ipam.go
@@ -355,16 +355,22 @@ func NormalizeIP(ip net.IP, networkName string) string {
 	return normalizedIP
 }
 
-// TODO: what's the best way to discover the node name? this should work in both controller pod and whereabouts host process
-func getNodeName() (string, error) {
+// getNodeName prefers an OS env var of NODENAME, or, uses a file named ./nodename in the whereabouts configuration path.
+func getNodeName(ipam *KubernetesIPAM) (string, error) {
+
 	envName := os.Getenv("NODENAME")
 	if envName != "" {
 		return strings.TrimSpace(envName), nil
 	}
-	file, err := os.Open("/etc/hostname")
+
+	nodeNamePath := fmt.Sprintf("%s/%s", ipam.Config.ConfigurationPath, "nodename")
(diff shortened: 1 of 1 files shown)


bpickard22:
ran e2e locally and all 27 of them passed @dougbtv

pkg/storage/kubernetes/ipam.go
```diff
@@ -356,15 +356,21 @@ func NormalizeIP(ip net.IP, networkName string) string {
 }
 
 // TODO: what's the best way to discover the node name? this should work in both controller pod and whereabouts host process
```
dougbtv, line 358: Let's update this line with:

> getNodeName prefers an OS env var of NODENAME, or, uses a file named ./nodename in the whereabouts configuration path.

pkg/storage/kubernetes/ipam.go
```diff
@@ -356,15 +356,21 @@ func NormalizeIP(ip net.IP, networkName string) string {
...
-	file, err := os.Open("/etc/hostname")
+
+	nodeNamePath := fmt.Sprintf("%s/%s", ipam.Config.ConfigurationPath, "nodename")
+	file, err := os.Open(nodeNamePath)
 	if err != nil {
-		logging.Errorf("Error opening file /etc/hostname: %v", err)
-		return "", err
+		file, err = os.Open("/etc/hostname")
+		if err != nil {
+			logging.Errorf("Error opening nodeNameFile, defaulted to /etc/hostname: %v", err)
```
dougbtv, line 371: let's change this to: 

> logging.Errorf("Could not determine nodename and could not open /etc/hostname: %v", err)

dougbtv (changes requested):
just some nits to clean up, thanks for this one Ben

dougbtv (approved):
lgtm! appreciate it Ben

dougbtv:
e2e test is flaking in the same spots. it's unrelated. thanks Ben, appreciate you spiffing it up.

Issue #575, opened 2025-03-04
[BUG] Infinite loop in `install-cni.sh` after 0.8.0


lubronzhan:
**Describe the bug**
This was added after 0.8.0 was cut https://github.com/k8snetworkplumbingwg/whereabouts/pull/495
I assume this change was to mimic what openshift was doing to workaround the token not refreshing issue here: https://github.com/openshift/multus-cni/blob/main/images/entrypoint.sh#L502-L508

The problem with whereabouts is that the dockerfile doesn't run `entrypoint.sh` directly, it runs install-cni.sh then ip-control-loop.
```
      - name: whereabouts
        command: [ "/bin/sh" ]
        args:
          - -c
          - >
            SLEEP=false /install-cni.sh &&
            /ip-control-loop -log-level debug
        image: ghcr.io/k8snetworkplumbingwg/whereabouts:latest
```

If the install-cni.sh never exit, ip-controller-loop will never be executed.

**Expected behavior**
Have a background process that update the content of the kubeconfig when running the ip-control-loop.
Or get rid of the kubeconfig in the whereabouts config file, since wherebouts ip-control-loop use client-go InClusterConfig to get service account's token and cert for kubeconfig.

**To Reproduce**
1. 
2. 
3.

**Environment**:
- Whereabouts version : master branch after 0.8.0
- Kubernetes version (use `kubectl version`): N/A
- Network-attachment-definition: N/A
- Whereabouts configuration (on the host): N/A
- OS (e.g. from /etc/os-release): N/A
- Kernel (e.g. `uname -a`): N/A
- Others: N/A

**Additional info / context**
That means the e2e test is not testing enough functionality
https://github.com/k8snetworkplumbingwg/whereabouts/blob/master/hack/e2e-setup-kind-cluster.sh

mlguerrero12:
Thanks @lubronzhan. I was investigating this issue as well.

I will review your PR.

c.c. @bpickard22 (https://github.com/k8snetworkplumbingwg/whereabouts/pull/495)

PR #577, opened 2025-03-05
Run token-watcher.sh in the background to refresh kubeconfig

lubronzhan:
With current script, install-cni.sh will stuck in an infinite loop, not continue to ip-control-loop.
This PR extract the loop into a background process `token-watcher.sh`

Example whereabouts log
```
kubo@OuN5Mh8kyHu3u:~$ kubectl logs -n kube-system -f whereabouts-hq9fs
Done configuring CNI.  Sleep=false
Sleep and Watching for service account token and CA file changes...
2025-03-05T07:23:58Z [debug] Filtering pods with filter key 'spec.nodeName' and filter value 'wl-calico-controlplane-spbtt-7qzd7'
2025-03-05T07:23:58Z [verbose] pod controller created
2025-03-05T07:23:58Z [verbose] Starting informer factories ...
2025-03-05T07:23:58Z [verbose] Informer factories started
2025-03-05T07:23:58Z [verbose] starting network controller
2025-03-05T07:23:59Z [verbose] using expression: 30 4 * *
```

Fixes #https://github.com/k8snetworkplumbingwg/whereabouts/issues/575

Other option would be like adding token refreshing using golang, like https://github.com/projectcalico/calico/pull/5910/files#diff-3bc859f09da6edba95b02904c6c5da879513d9d6e4c87e9c35ebc97b1afad6e6
Or rearchitect to thick model, having the shim binary talking to the whereabouts daemon through socket, instead of directly talking to apiserver

### DIFF
--- a/script/install-cni.sh
+++ b/script/install-cni.sh
@@ -10,149 +10,26 @@ set -u -e
 #
 #SPDX-License-Identifier: Apache-2.0
 
-CNI_BIN_DIR=${CNI_BIN_DIR:-"/host/opt/cni/bin/"}
-WHEREABOUTS_KUBECONFIG_FILE_HOST=${WHEREABOUTS_KUBECONFIG_FILE_HOST:-"/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig"}
-CNI_CONF_DIR=${CNI_CONF_DIR:-"/host/etc/cni/net.d"}
-WHEREABOUTS_RECONCILER_CRON=${WHEREABOUTS_RECONCILER_CRON:-30 4 * * *}
-
-# Make a whereabouts.d directory (for our kubeconfig)
-
-mkdir -p $CNI_CONF_DIR/whereabouts.d
-WHEREABOUTS_KUBECONFIG=$CNI_CONF_DIR/whereabouts.d/whereabouts.kubeconfig
-WHEREABOUTS_CONF_FILE=$CNI_CONF_DIR/whereabouts.d/whereabouts.conf 
-WHEREABOUTS_KUBECONFIG_LITERAL=$(echo "$WHEREABOUTS_KUBECONFIG" | sed -e s'|/host||')
-
-# ------------------------------- Generate a "kube-config"
-SERVICE_ACCOUNT_PATH=/var/run/secrets/kubernetes.io/serviceaccount
-KUBE_CA_FILE=${KUBE_CA_FILE:-$SERVICE_ACCOUNT_PATH/ca.crt}
-SERVICE_ACCOUNT_TOKEN=$(cat $SERVICE_ACCOUNT_PATH/token)
-SERVICE_ACCOUNT_TOKEN_PATH=$SERVICE_ACCOUNT_PATH/token
-SKIP_TLS_VERIFY=${SKIP_TLS_VERIFY:-false}
-
-LAST_SERVICEACCOUNT_MD5SUM=""
-LAST_KUBE_CA_FILE_MD5SUM=""
+source lib.sh
 
 # Setup our logging routines
 
-function log()
-{
-    echo "$(date --iso-8601=seconds) ${1}"
-}
 
-function error()
-{
-    log "ERR:  {$1}"
-}
-
-function warn()
-{
-    log "WARN: {$1}"
-}
-
-
-function generateKubeConfig {
-  # Check if we're running as a k8s pod.
-if [ -f "$SERVICE_ACCOUNT_PATH/token" ]; then
-  # We're running as a k8d pod - expect some variables.
-  if [ -z ${KUBERNETES_SERVICE_HOST} ]; then
-    error "KUBERNETES_SERVICE_HOST not set"; exit 1;
-  fi
-  if [ -z ${KUBERNETES_SERVICE_PORT} ]; then
-    error "KUBERNETES_SERVICE_PORT not set"; exit 1;
-  fi
-
-  if [ "$SKIP_TLS_VERIFY" == "true" ]; then
-    TLS_CFG="insecure-skip-tls-verify: true"
-  elif [ -f "$KUBE_CA_FILE" ]; then
-    TLS_CFG="certificate-authority-data: $(cat $KUBE_CA_FILE | base64 | tr -d '\n')"
-  fi
-
-  # Kubernetes service address must be wrapped if it is IPv6 address
-  KUBERNETES_SERVICE_HOST_WRAP=$KUBERNETES_SERVICE_HOST
-  if [ "$KUBERNETES_SERVICE_HOST_WRAP" != "${KUBERNETES_SERVICE_HOST_WRAP#*:[0-9a-fA-F]}" ]; then
-    KUBERNETES_SERVICE_HOST_WRAP=\[$KUBERNETES_SERVICE_HOST_WRAP\]
-  fi
-
-  # Write a kubeconfig file for the CNI plugin.  Do this
-  # to skip TLS verification for now.  We should eventually support
-  # writing more complete kubeconfig files. This is only used
-  # if the provided CNI network config references it.
-  touch $WHEREABOUTS_KUBECONFIG
-  chmod ${KUBECONFIG_MODE:-600} $WHEREABOUTS_KUBECONFIG
-  cat > $WHEREABOUTS_KUBECONFIG <<EOF
-# Kubeconfig file for the Whereabouts CNI plugin.
-apiVersion: v1
-kind: Config
-clusters:
-- name: local
-  cluster:
-    server: ${KUBERNETES_SERVICE_PROTOCOL:-https}://${KUBERNETES_SERVICE_HOST_WRAP}:${KUBERNETES_SERVICE_PORT}
-    $TLS_CFG
-users:
-- name: whereabouts
-  user:
-    token: "${SERVICE_ACCOUNT_TOKEN}"
-contexts:
-- name: whereabouts-context
-  context:
-    cluster: local
-    user: whereabouts
-    namespace: ${WHEREABOUTS_NAMESPACE}
-current-context: whereabouts-context
-EOF
-
-else
-  warn "Doesn't look like we're running in a kubernetes environment (no serviceaccount token)"
-fi
-
-}
-
+# -------------------Generate a "kube-config"
 generateKubeConfig
-
+export LAST_SERVICEACCOUNT_MD5SUM="$(get_token_md5sum)"
+export LAST_KUBE_CA_FILE_MD5SUM="$(get_ca_file_md5sum)"
 # ------------------ end Generate a "kube-config"
 
 # ----------------- Generate a whereabouts conf
-
-function generateWhereaboutsConf {
-
-  touch $WHEREABOUTS_CONF_FILE
-  chmod ${KUBECONFIG_MODE:-600} $WHEREABOUTS_CONF_FILE
-  cat > $WHEREABOUTS_CONF_FILE <<EOF
-{
-  "datastore": "kubernetes",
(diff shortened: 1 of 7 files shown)


script/token-watcher.sh
```diff
@@ -0,0 +1,19 @@
...
+
+source lib.sh
+
+echo "Sleep and Watching for service account token and CA file changes..."
+# enter sleep/watch loop
+while true; do
+  # Check the md5sum of the service account token and ca.
+  svcaccountsum=$(md5sum $SERVICE_ACCOUNT_TOKEN_PATH | awk '{print $1}')
+  casum=$(md5sum $KUBE_CA_FILE | awk '{print $1}')
+  if [ "$svcaccountsum" != "$LAST_SERVICEACCOUNT_MD5SUM" ] || [ "$casum" != "$LAST_KUBE_CA_FILE_MD5SUM" ]; then
```
mlguerrero12, line 13: I don't see where LAST_SERVICEACCOUNT_MD5SUM and LAST_KUBE_CA_FILE_MD5SUM are updated (also in current code)

script/token-watcher.sh
```diff
@@ -0,0 +1,19 @@
...
+    # log "Detected service account or CA file change, regenerating kubeconfig..."
+    generateKubeConfig
+  fi
+
+  sleep 1h
```
mlguerrero12, line 18: You changed it to 1h instead of 1s because the default duration of the token is 1h? What if we miss it just before the refresh? We will have authentication errors while this is sleeping

mlguerrero12:
I think this approach is good to fix what we currently have, but ideally, we should refresh it in code once we get a 401 error. Is that what Calico does?

script/token-watcher.sh
lubronzhan: Oh I remember the lifetime wrong. Changed it back

script/token-watcher.sh
lubronzhan: Good catch, updated. Blindly copied over

lubronzhan:
> I think this approach is good to fix what we currently have, but ideally, we should refresh it in code once we get a 401 error. Is that what Calico does?

Calico just rotate the token in kubeconfig if it is about to expire in `equal or less than 15 minutes`, if default ttl of the token is 1 hour. (buffer time is 1/4 of ttl).
Yeah ideally golang would be good. Now this is just a simple working version

script/install-cni.sh
```diff
@@ -10,150 +10,29 @@ set -u -e
...
-  warn "Doesn't look like we're running in a kubernetes environment (no serviceaccount token)"
-fi
-
-}
 
+# -------------------Generate a "kube-config"
 generateKubeConfig
-
+export LAST_SERVICEACCOUNT_MD5SUM="$(get_token_md5sum)"
+if ! [ "$SKIP_TLS_VERIFY" == "true" ]; then
```
mlguerrero12, line 21: I think it's fine to not check SKIP_TLS_VERIFY here and simply export LAST_KUBE_CA_FILE_MD5SUM given that SKIP_TLS_VERIFY is used in token-watcher.sh

script/token-watcher.sh
```diff
@@ -0,0 +1,25 @@
...
+
+set -u -e
+
+source lib.sh
+
+echo "Sleep and Watching for service account token and CA file changes..."
+# enter sleep/watch loop
+while true; do
+  # Check the md5sum of the service account token and ca.
+  svcaccountsum=$(md5sum $SERVICE_ACCOUNT_TOKEN_PATH | awk '{print $1}')
```
mlguerrero12, line 11: why not using the functions get_token_md5sum and get_ca_file_md5sum here?

script/token-watcher.sh
```diff
@@ -0,0 +1,25 @@
...
+  if [ -f "$KUBE_CA_FILE" ]; then
+    casum=$(md5sum $KUBE_CA_FILE | awk '{print $1}')
+  fi
+  if [ "$svcaccountsum" != "$LAST_SERVICEACCOUNT_MD5SUM" ] || ! [ "$SKIP_TLS_VERIFY" == "true" ] && [ "$casum" != "$LAST_KUBE_CA_FILE_MD5SUM" ]; then
+    log "Detected service account or CA file change, regenerating kubeconfig..."
+    generateKubeConfig
+    LAST_SERVICEACCOUNT_MD5SUM="$(get_token_md5sum)"
```
mlguerrero12, line 18: I was expecting something like
LAST_SERVICEACCOUNT_MD5SUM = svcaccountsum
LAST_KUBE_CA_FILE_MD5SUM = casum

it seems more natural to do that instead of calling the functions again

script/install-cni.sh
lubronzhan: Cool

script/token-watcher.sh
lubronzhan: Good catch

script/token-watcher.sh
lubronzhan: Yeah good catch

script/token-watcher.sh
```diff
@@ -0,0 +1,23 @@
...
+set -u -e
+
+source lib.sh
+
+echo "Sleep and Watching for service account token and CA file changes..."
+# enter sleep/watch loop
+while true; do
+  # Check the md5sum of the service account token and ca.
+  svcaccountsum="$(get_token_md5sum)"
+  if [ -f "$KUBE_CA_FILE" ]; then
```
mlguerrero12, line 12: please remove this checks since it is already in get_ca_file_md5sum

script/lib.sh
```diff
@@ -0,0 +1,121 @@
...
+SERVICE_ACCOUNT_TOKEN=$(cat $SERVICE_ACCOUNT_PATH/token)
+SERVICE_ACCOUNT_TOKEN_PATH=$SERVICE_ACCOUNT_PATH/token
+SKIP_TLS_VERIFY=${SKIP_TLS_VERIFY:-false}
+
+LAST_SERVICEACCOUNT_MD5SUM=""
+LAST_KUBE_CA_FILE_MD5SUM=""
+
+function log()
+{
+    echo "$(date --iso-8601=seconds) ${1}"
```
mlguerrero12, line 24: Tested your pr and I'm getting an error due to this
date: unrecognized option: iso-8601=seconds
BusyBox v1.36.1 (2023-11-07 18:53:09 UTC) multi-call binary.

Usage: date [OPTIONS] [+FMT] [[-s] TIME]

Display time (using +FMT), or set time

	-u		Work in UTC (don't convert to local time)
	[-s] TIME	Set time to TIME
	-d TIME		Display TIME, not 'now'
	-D FMT		FMT (strptime format) for -s/-d TIME conversion
	-r FILE		Display last modification time of FILE
	-R		Output RFC-2822 date
	-I[SPEC]	Output ISO-8601 date
			SPEC=date (default), hours, minutes, seconds or ns

Recognized TIME formats:
	@seconds_since_1970
	hh:mm[:ss]
	[YYYY.]MM.DD-hh:mm[:ss]
	YYYY-MM-DD hh:mm[:ss]
	[[[[[YY]YY]MM]DD]hh]mm[.ss]
	'date TIME' form accepts MMDDhhmm[[YY]YY][.ss] instead

Do you see it as well? Fix could be

    echo "$(date -Iseconds) ${1}"

script/lib.sh
```diff
@@ -0,0 +1,121 @@
...
+WHEREABOUTS_CONF_FILE=$CNI_CONF_DIR/whereabouts.d/whereabouts.conf
+WHEREABOUTS_KUBECONFIG_LITERAL=$(echo "$WHEREABOUTS_KUBECONFIG" | sed -e s'|/host||')
+
+SERVICE_ACCOUNT_PATH=/var/run/secrets/kubernetes.io/serviceaccount
+KUBE_CA_FILE=${KUBE_CA_FILE:-$SERVICE_ACCOUNT_PATH/ca.crt}
+SERVICE_ACCOUNT_TOKEN=$(cat $SERVICE_ACCOUNT_PATH/token)
+SERVICE_ACCOUNT_TOKEN_PATH=$SERVICE_ACCOUNT_PATH/token
+SKIP_TLS_VERIFY=${SKIP_TLS_VERIFY:-false}
+
+LAST_SERVICEACCOUNT_MD5SUM=""
```
mlguerrero12, line 19: I see you have in install-cni.sh

export LAST_SERVICEACCOUNT_MD5SUM="$(get_token_md5sum)"
export LAST_KUBE_CA_FILE_MD5SUM="$(get_ca_file_md5sum)"

but when you source lib.sh in token-watcher.sh, these values are empty again due to these definition. Here, it should be the current value or empty is not defined. 

LAST_SERVICEACCOUNT_MD5SUM="${LAST_SERVICEACCOUNT_MD5SUM:-}"

Regardless of this, the commands are run in different subshells, therefore, the export in install-cni.sh does not really work. These env vars are not shared. There are many way to solve this, one is sourcing the scripts

          - |
            SLEEP=false
            source install-cni.sh
            source token-watcher.sh &

Gave it a try and it works but feel free to propose something else

lubronzhan:
Interesting, I decoded the token, actually the ttl is 1 year
```
/ # cat /var/run/secrets/kubernetes.io/serviceaccount/token
eyJhbGciOiJSUzI1NiIsImtpZCI6Ik9NZjBLWXlLSHlIRDhZWmx0SEFkNURxOE5YTVU3bDVjdi1fdjJ4bkJfZnMifQ.eyJhdWQiOlsiaHR0cHM6Ly9rdWJlcm5ldGVzLmRlZmF1bHQuc3ZjLmNsdXN0ZXIubG9jYWwiXSwiZXhwIjoxNzczNTQ3NDYxLCJpYXQiOjE3NDIwMTE0NjEsImlzcyI6Imh0dHBzOi8va3ViZXJuZXRlcy5kZWZhdWx0LnN2Yy5jbHVzdGVyLmxvY2FsIiwia3ViZXJuZXRlcy5pbyI6eyJuYW1lc3BhY2UiOiJrdWJlLXN5c3RlbSIsInBvZCI6eyJuYW1lIjoid2hlcmVhYm91dHMtNnQ1Nm4iLCJ1aWQiOiI3MWFmNjFlMS0zZDNhLTQzNWEtYTUxZS01NWFkMDhhOTQ3NWMifSwic2VydmljZWFjY291bnQiOnsibmFtZSI6IndoZXJlYWJvdXRzIiwidWlkIjoiNTE2OTgyZTItZDYxNC00NTY5LTg0YjgtZDNjNzliN2FmNGYxIn0sIndhcm5hZnRlciI6MTc0MjAxNTA2OH0sIm5iZiI6MTc0MjAxMTQ2MSwic3ViIjoic3lzdGVtOnNlcnZpY2VhY2NvdW50Omt1YmUtc3lzdGVtOndoZXJlYWJvdXRzIn0.ff-UqSVH2RKpooWOiceaWGXGbDyh1JbcCPlwq36Eau0ffR6x2x2G99heHd6tnLcNIX6_GFk0PERejeYYlWN74MQtuh16TKtRI0cNKcEngyBZg_MvPxyZrbPEXl8mWuBH52hR8uuqwoNzoz30-jkQC_2ZxSNR1EQsuKhwEQpHawr2fRubB6JREscGsl5tkccYOQdMucIszKf-B6VPGcapf-eZYDuZTe0xch0ALkBJVE76-TLwpCWcq5QvuWQiEXHFKF6Y_BnlhQBja1e8tqUjYnPWkVMrDCcXnsGtBQHTxxh4yI0mDxakr4fuizoldve-ZnSzvNZg6F2KfksIjLTccw
```
image

mlguerrero12:
LGTM

Thanks @lubronzhan. Before I merge this, please organize the commits. Each one should have a relevant title and description. You could also squash them and have a single one.

PR #331, opened 2023-05-18
Add dependabot config

sebrandon1:
This PR adds a dependabot config that enables:
- go.mod updates
- Dockerfile updates
- github-actions updates

PR #589, opened 2025-04-01
Improve dependabot config with groups

sebrandon1:
Follow up to my original PR: #331 

Improves the `go.mod` dependabot settings by grouping everything k8s related into its own PR.  All other dependencies will be in their own group.


mlguerrero12 (approved):
LGTM

PR #592, opened 2025-04-02
Remove duplicate app label from helm chart

### LINKED ISSUE #591: [BUG] Duplicate app label in helm template
**Describe the bug**
Unable to use the helm chart using fluxcd because the `app` label on resources is set twice in the templates.

**Expected behavior**
Only one `app` label should be set

**To Reproduce**
1. `helm template whereabouts oci://ghcr.io/k8snetworkplumbingwg/whereabouts-chart`
2. Observe `ServiceAccount` and `DaemonSet` `metadata.labels` contains two entries for `app`
3. Installing with fluxcd results in:
```text
Helm install failed for release kube-system/whereabouts with chart whereabouts-chart@0.9.0+1c894040bcf0: error while running post render on files: map[string]interface {}(nil): yaml: unmarshal errors: line 10: mapping key "app" already defined at line 8
```

**Environment**:
- Whereabouts version : v0.9.0
- Kubernetes version (use `kubectl version`): Client Version: v1.32.2, Kustomize Version: v5.5.0, Server Version: v1.32.3
- Others: N/A

cyanidium:
Only one `app` label should be set on kubernetes resources. fluxcd will refuse to install the chart with duplicate app labels. This removes a hardcoded app label and leaves the label based on the chart name in place.

Fixes #591

### DIFF
--- a/deployment/whereabouts-chart/templates/_helpers.tpl
+++ b/deployment/whereabouts-chart/templates/_helpers.tpl
@@ -45,7 +45,6 @@ Create chart name and version as used by the chart label.
 Common labels
 */}}
 {{- define "whereabouts.labels" -}}
-app: whereabouts
 helm.sh/chart: {{ include "whereabouts.chart" . }}
 {{ include "whereabouts.selectorLabels" . }}
 {{- if .Chart.AppVersion }}


deployment/whereabouts-chart/templates/_helpers.tpl
```diff
@@ -45,7 +45,6 @@ Create chart name and version as used by the chart label.
 Common labels
 */}}
 {{- define "whereabouts.labels" -}}
-app: whereabouts
```
mlguerrero12, line 48: I see that the app is defined below in selectorLabels, and the value is Chart.name, which is whereabouts-chart. I'm wondering if we should change this to just whereabouts instead. It’s just a cosmetic change.

cyanidium: It is unusual to have "chart" in the app name, yes.

mlguerrero12: Let's tackle that in another pr

mlguerrero12 (approved):
LGTM

PR #598, opened 2025-04-08
Reconciler: skip failed or succeeded Pods in live Pod list

xagent003:
Jobs remain in Failed or Succeeded state - they are not Running nor Pending. As a result the ip-reconciler is not cleaning up their IPs.

We should consider terminal Pods as gone, and not included in the live whereabouts Pods map. This way the reconciler will clean them up

### DIFF
--- a/pkg/reconciler/wrappedPod.go
+++ b/pkg/reconciler/wrappedPod.go
@@ -47,7 +47,7 @@ func indexPods(livePodList []v1.Pod, whereaboutsPodNames map[string]void) map[st
 			continue
 		}
 
-		if isPodMarkedForDeletion(pod.Status.Conditions) {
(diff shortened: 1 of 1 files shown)


xagent003:
@dougbtv @maiqueb can you please take a look at this as well? diff is quite large due to auto-generated code but the actual change is only in wrappedPod.go

mlguerrero12:
Just introduce the changes in wrappedPod.go. Why do you need the other changes (autogenerated changes)?

xagent003:
> Just introduce the changes in wrappedPod.go. Why do you need the other changes (autogenerated changes)?

Hi @mlguerrero12 sorry for the long delay. I undid the cherry-pick and just added the change to wrappedPod.go. Can you or @dougbtv please take a look now?

PR #599, opened 2025-04-09
Reconciler and status check optimizations

xagent003:
A few optimization fixes:

 - return if leader election fails
 - no need for the costly list IPpools call at start of each IPAM to "test" API server connectivity. We get the IPpools anyways and use client-go leader election.
 - refactored the isOrphanedIP loop in reconciler and also noticed the logic was backwards for the return value. Fixed that and we can just do a map lookup rather the iterate through the entire livePod list which can be huge

### DIFF
--- a/pkg/reconciler/iploop.go
+++ b/pkg/reconciler/iploop.go
@@ -75,8 +75,7 @@ func (rl *ReconcileLooper) findOrphanedIPsPerPool(ipPools []storage.IPPool) erro
 				_ = logging.Errorf("pod ref missing for Allocations: %s", ipReservation)
 				continue
 			}
-			if !rl.isOrphanedIP(ipReservation.PodRef, ipReservation.IP.String()) {
-				logging.Debugf("pod ref %s is not listed in the live pods list", ipReservation.PodRef)
+			if rl.isOrphanedIP(ipReservation.PodRef, ipReservation.IP.String()) {
 				orphanIP.Allocations = append(orphanIP.Allocations, ipReservation)
 			}
 		}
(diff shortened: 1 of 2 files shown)


xagent003:
@dougbtv @maiqueb Can you please take a look?

pkg/reconciler/iploop.go
```diff
@@ -88,49 +88,55 @@ func (rl *ReconcileLooper) findOrphanedIPsPerPool(ipPools []storage.IPPool) erro
...
+					logging.Debugf("Found IP on refreshed pending pod, not orphaned")
+					return false
 				}
-				isFound = isIpOnPod(podToMatch, podRef, ip)
+				time.Sleep(time.Duration(500) * time.Millisecond)
 			}
+		}
 
-			return isFound
+		isIPFoundOnPod = isIpOnPod(podToMatch, podRef, ip)
```
mlguerrero12, line 128: I would leave as it was
`isFound = isIpOnPod(podToMatch, podRef, ip)`

at at the end return

`return !isIPFoundOnPod` instead of 

`
	// IP is not found on pod, so it IS orphaned
	return true
`

pkg/reconciler/iploop.go
```diff
@@ -88,49 +88,55 @@ func (rl *ReconcileLooper) findOrphanedIPsPerPool(ipPools []storage.IPPool) erro
...
+		podToMatch := &livePod
+		retries := 0
+
+		logging.Debugf("Re-fetching Pending Pod: %s IP-to-match: %s", podRef, ip)
+
+		for retries < storage.PodRefreshRetries {
+			retries++
+			podToMatch = rl.refreshPod(podRef)
+			if podToMatch == nil {
+				logging.Debugf("Pod refresh returned nil, IP is orphaned")
```
mlguerrero12, line 113: let's just continue here so that it tries to get the pod again. It might be a temporal issue.

pkg/reconciler/iploop.go
```diff
@@ -88,49 +88,55 @@ func (rl *ReconcileLooper) findOrphanedIPsPerPool(ipPools []storage.IPPool) erro
...
+				return true
+			} else if podToMatch.phase != v1.PodPending {
```
mlguerrero12, line 115: let's remove some `else` statements 

like 

```
					if podToMatch.phase != v1.PodPending {
						logging.Debugf("Pending Pod is now in phase: %s", podToMatch.phase)
						break
					}
					isFound = isIpOnPod(podToMatch, podRef, ip)
					// Short-circuit - Pending Pod may have IP now
					if isFound {
						logging.Debugf("Pod now has IP annotation while in Pending")
						return true
					}
					time.Sleep(time.Duration(250) * time.Millisecond)
```

mlguerrero12 (commented):
thanks @xagent003, posted some comments

pkg/reconciler/iploop.go
xagent003: Fixed

PR #605, opened 2025-04-23
Fix chart and source command in scripts

thomasferrandiz:
- Align the chart with the daemonset manifest
- Fix source command in scripts: In busybox-based images, source does not look in the current directory so an absolute path is needed.

Fixes #

### DIFF
--- a/deployment/whereabouts-chart/templates/node-slice-controller.yaml
+++ b/deployment/whereabouts-chart/templates/node-slice-controller.yaml
@@ -1,3 +1,4 @@
+# Dont'forget to update doc/crds/node-slice-controller.yaml as well
 {{- if .Values.nodeSliceController.enabled }}
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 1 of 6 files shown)


mlguerrero12:
is the deployment with helm charts failing without the /lib? It uses the same image

thomasferrandiz:
No the chart just needed an update to be aligned with the standard manifest.

The `/lib` change is to fix is an issue when repackaging downstream using a busybox-based image.
The `source` command included with busybox needs an absolute path to work properly.

mlguerrero12:
yes, we forgot to also change the daemonset in the charts when we modified the one under the `doc` dir. We should have a comment in both files to not forget next time.

I see, so, it is for repacking downstream. LGTM

Would you mind adding the comment I mentioned?

deployment/whereabouts-chart/templates/node-slice-controller.yaml
```diff
@@ -1,3 +1,4 @@
+# Dont'forget to doc/crds/node-slice-controller.yaml as well
```
bpickard22, line 1: nit: add update

thomasferrandiz: done

mlguerrero12 (approved):

PR #610, opened 2025-05-05
[wip] Add metrics pkg

bpickard22:
enables metrics server deployment for whereabouts to collect prom metrics and enable pprof


bpickard22:
closing due to staleness - will get back to this if i have cycles

Issue #518, opened 2024-11-26
[BUG] fast ranges not working out of the box


dougbtv:
**Describe the bug**
Using fast ranges results

**Expected behavior**
The pod should come up, however, it fails with:

```
[...]

PR #614, opened 2025-05-19
🐛 Fix the namespace of NAD in NodeSlicePool's ownerRef

lubronzhan:
Right now it's hardcoded to whereabout's namespace, but NAD might be deployed in different namespace. If deployed in different namespace, NodeSlicePool will be garbage collected immediate after creation, and following logic in Whereabout will fail and report nil pointer.
Example in comment in this https://github.com/k8snetworkplumbingwg/whereabouts/issues/518#issuecomment-2887775023

Fixes # https://github.com/k8snetworkplumbingwg/whereabouts/issues/518

### DIFF
--- a/pkg/node-controller/controller.go
+++ b/pkg/node-controller/controller.go
@@ -367,7 +367,7 @@ func (c *Controller) syncHandler(ctx context.Context, key string) error {
 			},
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      getSliceName(ipamConf),
-				Namespace: c.whereaboutsNamespace,
+				Namespace: namespace,
 				OwnerReferences: []metav1.OwnerReference{
 					*metav1.NewControllerRef(nad, cncfV1.SchemeGroupVersion.WithKind("NetworkAttachmentDefinition")),
 				},


lubronzhan:
Ohk jsut found out can't set ownerRef to a resource in different namespace. So maybe this hardcoded namespace was intentional. 

But NAD could be presented in any namespace in the doc https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/configuration.md#namespace-isolation
Should we then allow user to create NodeSlicePool be defined in different namespace as well then

lubronzhan:
But is NodeSlicePool supposed to be 1 map to 1 cluster? Then NAD should be in the same namespace, then how does the pod in different namespace reference this NAD

xagent003:
Pod doesn't need to be in the same namespace as a NAD. The Pod just references the NAD via multus annotation <namespace>/<nad name>

We keep all NADs in one namespace, after all they generally attach to physical networks

lubronzhan:
IC, in that way, should we have a webhook to reject NAD created in other namespace. Otherwise it will just cause nil pointer exception mentioned in https://github.com/k8snetworkplumbingwg/whereabouts/issues/518#issuecomment-2887775023
Or we should at least log a warning in the log?

PR #615, opened 2025-05-27
Add priorityClassName support to Helm chart

kuzaxak:

To make sure we can mark daemonsets pods as first priority to be scheduled before the apps we need to support `priorityClassName` field.

By default, it won't be added.


kuttattz:
@mlguerrero12 
Can you please let me know when will be the next release with this fix?

PR #619, opened 2025-06-17
add stalebot to the repository

thomasferrandiz:

This PR adds the configuration for a stale bot to automatically close issues after 1 year.


mlguerrero12:
Shouldn´t we do the same for PRs? I would believe it is even more relevant there. WDYT?

PR #624, opened 2025-07-14
Prevent accidental IP deallocation in statefulsets

mlguerrero12:
This commit addresses a potential issue where IP allocations could be accidentally deleted when new pods with the same name and namespace are created in statefulset scenarios. The logic now correctly checks if a pod with the matching name and namespace exists and is not marked for deletion. If such a pod exists, the deallocation is skipped, preventing conflicts. The commit also includes clarified logging messages for improved debugging.


pkg/controlloop/pod.go
```diff
@@ -222,6 +223,21 @@ func (pc *PodController) garbageCollectPodIPs(pod *v1.Pod) error {
 		for _, pool := range pools {
 			for allocationIndex, allocation := range pool.Spec.Allocations {
 				if allocation.PodRef == podID(podNamespace, podName) {
+					logging.Verbosef("Found an existing allocation: %+v", allocation)
+
+					// The allocation could belong to a new pod with the same name and namespace. Stateful set scenarios.
+					// The previous pod should be gone by the time a pod deletion event is received.
+					if newPod, err := pc.k8sClient.CoreV1().Pods(podNamespace).Get(context.TODO(), podName, metav1.GetOptions{}); err == nil {
```
bpickard22, line 230: should we check if the pod is owned by a statefulset so we dont break the behavior for non stateful deployments?

mlguerrero12: This doesn't apply to non stateful deployments. Pod name is different.

PR #649, opened 2025-11-14
single-ip allocation and network/broadcast address inclusion

kruftik:

A feature of single-IP allocation (from a list of provided IP pools/ranges) is introduced. Such a feature allows to configure as many IP pools as we want and have only one IP assigned from whichever pool with spare "capacity".

In addition to, for cases when address pools are just parts of some larger network, split into a list of smaller networks, it's possible to include the first (network) and the last (broadcast) addresses of a CIDR to the allocatable range.

## Summary by CodeRabbit

* **New Features**
  * Single‑IP mode and options to include network/broadcast addresses in IPAM ranges.
  * Improved support for multiple IP pools, dual‑stack and named‑range scenarios.

* **Refactor**
  * Introduced a unified Pool abstraction for IP range handling.
  * Standardized operation‑type semantics for allocation/deallocation and richer per‑range controls.

* **Tests**
  * Expanded end‑to‑end and unit test coverage for pool, single‑IP and multi‑pool behaviors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ build:
 	hack/build-go.sh
 
 docker-build:
-	$(OCI_BIN) build -t ${IMAGE_REGISTRY}/${IMAGE_NAME}:${IMAGE_TAG} -f Dockerfile --platform linux/amd64 .
+	$(OCI_BIN) build -t ${IMAGE_REGISTRY}/${IMAGE_NAME}:${IMAGE_TAG} -f Dockerfile .
 
 generate-api:
 	hack/verify-codegen.sh
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ build:
 	hack/build-go.sh
 
 docker-build:
-	$(OCI_BIN) build -t ${IMAGE_REGISTRY}/${IMAGE_NAME}:${IMAGE_TAG} -f Dockerfile .
+	$(OCI_BIN) build -t ${IMAGE_REGISTRY}/${IMAGE_NAME}:${IMAGE_TAG} -f Dockerfile --platform linux/amd64 .
 
 generate-api:
 	hack/verify-codegen.sh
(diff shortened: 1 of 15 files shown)


alexferenets:
@maiqueb Hi!
Do you think you can take a look at the PR?

maiqueb:
There are conflicts that must be fixed.

maiqueb:
The commit history also needs to be simplified. We can't merge it like this.

maiqueb (changes requested):
I would like to see an enhancement proposal for this before we jump into the implementation. 

This will change the API, and I'd rather do it after considering options, and in a way that allows for future growth.

kruftik:
> I would like to see an enhancement proposal for this before we jump into the implementation.
> This will change the API, and I'd rather do it after considering options, and in a way that allows for future growth.

we discussed that change half an year ago, you can see my and your own comments above

maiqueb:
> > I would like to see an enhancement proposal for this before we jump into the implementation.
> > This will change the API, and I'd rather do it after considering options, and in a way that allows for future growth.
> 
> we discussed that change half an year ago, you can see my and your own comments above

yes, we have discussed this, and I bought the use case. I also said nothing about the API, but, given APIs live forever, and we won't be able to change it later on, I'd rather have an API that makes sense, rather than something that works, that's confusing / hard to work with / has no growth potential. Using booleans directly hints at the last item in the list.

I'm not saying the change is wrong - I am just saying there's plenty of ways this can implemented, and i want to hash them out, understand what's the best option, and see the best option implemented. Not something that works for you and me.

And, FWIW, please keep in mind my current professional objectives have driven me away from whereabouts. I do not actively maintain this on my employer's time. I am looking at this only from a community perspective, and trying to keep the community alive.

fabrizio2210:
If you want to discuss a proposal for the single-IP feature, I created https://github.com/k8snetworkplumbingwg/whereabouts/pull/738

PR #672, opened 2026-02-16
Add leader election to reconciler

### LINKED ISSUE #386: Memory limit of 200Mi for the whereabouts daemon-set inadequate for large clusters
With the 200Mi memory limit for the whereabouts daemon-set the POD is getting OOMKilled in large clusters (100+ workers). 
We locally tested with limit set to 400Mi and the issue is not seen.

Raising this ticket to consider that change.

AMacedoP:
Adds leader election to the reconciler execution to ensure that it runs on only one pod of the whereabouts daemonset.

Fixes #386.

### DIFF
--- a/cmd/controlloop/controlloop.go
+++ b/cmd/controlloop/controlloop.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"os"
@@ -10,10 +11,13 @@ import (
 	"github.com/fsnotify/fsnotify"
 	"github.com/go-co-op/gocron/v2"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"k8s.io/client-go/tools/record"
 
 	nadclient "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/clientset/versioned"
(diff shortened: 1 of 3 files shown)


fmuyassarov:
Hi @thomasferrandiz, would you be able to take a look?

I was about to report an issue related to OOM errors occurring in large clusters (around 150 nodes). It seems that all pods in the cluster are being listed simultaneously and loaded into memory at the same time, which is causing the OOM condition and it seems we are not the only one experiencing the same issue with whereabouts.

fmuyassarov:
Ping @thomasferrandiz

thomasferrandiz (approved):
LGTM

thomasferrandiz:
@bpickard22 could you have a look as well?

mlguerrero12:
Why not running the cron scheduler as a Deployment instead? This Daemonset has a pod controller that needs to run in every node but the cron scheduler should run in a Deployment.

AMacedoP:
@mlguerrero12 I really like your idea because one downside of running the reconciler in the daemonset is that finding the pod with the logs is difficult. If all the other maintainers are ok with it I can create a new PR creating a new Deployment for the reconciler and running it with replicas=1

mlguerrero12:
Indeed, this will also improve troubleshooting.

Any objections @thomasferrandiz? 

c.c. @karampok FYI, we've discussed this in the past

karampok:
Hi!
I did not have the opportunity to deep dive in the codebase so my opinion might be off.

For this PR: adding this leader election in daemonset, it looks like we implement deployment. So we should challenge that. Also I would like to have some testing on the leader election implementation, but not sure if the repo/code can  easily support that.

@mlguerrero12 doing a deployment set with replica 1* that inside does a cronjob sounds like we need a cronjob from k8s. Wdyt? (https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/)

* one question is if we care about fast failover, which is probably not

mlguerrero12:
Yeah, this PR shouldn't go in in my opinion. The cronjob workload can be explored but the fastest/cleanest solution right now is extracting the logic around the cron schedule that we know it works in production setups, and creating a deployment for it.

@AMacedoP, If you have time, please create a PR for this. I'll be happy to review it.

thomasferrandiz:
> Indeed, this will also improve troubleshooting.
> 
> Any objections @thomasferrandiz?
> 
> c.c. @karampok FYI, we've discussed this in the past

no objections from me.
The setup will be simpler this way.

fmuyassarov:
Hi @mlguerrero12 @thomasferrandiz 
Are you suggesting to replace the all the manifests and references in the documentation to be CronJob based or shall that be in parallel to what exists today?

fmuyassarov:
Hi @mlguerrero12 @thomasferrandiz
Agreed on the CronJob approach. Just to note that this approach still does require a some amount code changes anyways alongside the manifest work. Because, all current binaries are long-running processes, ip-control-loop runs both the pod controller and an internal gocron scheduler indefinitely. A cronJob container needs a command that runs once and exits. So, we would need `cmd/reconciler/main.go` that calls the existing `pkg/reconciler/ip.go` logic once and exits  with the appropriate exit code.  We also need to remove the internal gocron scheduler from ip-control-loop, otherwise it will still fire the reconciler internally on its own schedule in addition to the CronJob.

Is my understanding correct and are you okay with the following plan? If yes, I'm happy to file a changes.

karampok:
>  but the fastest/cleanest solution right now is extracting the logic around the cron schedule that we know it works in production setups, and creating a deployment for it.

My understanding is that @mlguerrero12 prefers having a deployment, but better he confirms first.

mlguerrero12:
Yes, move the internal gocron scheduler to a Deployment

fmuyassarov:
@mlguerrero12 @thomasferrandiz, 

I've filed #700  as suggested above. Can you please take a look?

AMacedoP:
Thanks @fmuyassarov for working on the new proposal because I didn't have time to work on it. I'll close this PR

PR #724, opened 2026-06-29
Fix node watch predicate to avoid reconcile storms

MaxRink:
Closes #238


MaxRink:
Opened by mistake against upstream

PR #736, opened 2026-07-31
Fix reconciler cron ConfigMap reload race on first projection

smulje:
## What this PR does / why we need it:
Fixes a race where the IP reconciler cron schedule ConfigMap (`whereabouts-config`) is mounted successfully, but some reconciler pods keep using the default schedule (`30 4 * * *`) until the pod is restarted.

## Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): 

Special notes for your reviewer (optional):

## Root cause

Kubernetes ConfigMap projection updates the mount roughly as:

1. write a new timestamped directory
2. atomically swap the `..data` symlink
3. create/update the user-visible `config` symlink (`config` → `..data/config`)

The watcher previously treated only `CREATE` on `/cron-schedule/..data` as relevant, then immediately read `/cron-schedule/config`.

On **first-time** ConfigMap projection, kubelet can emit `CREATE` on `..data` **before** the `config` symlink exists. In that case:

1. `ReadFile("/cron-schedule/config")` fails
2. the code falls back to the flatfile default (`30 4 * * *`)
3. schedule compares equal to the current value → `no changes in schedule, nothing to do`
4. the later `CREATE /cron-schedule/config` event is ignored as `event not relevant`
5. the pod stays stuck on the default cron until restart

This matches the failure logs:

```text
could not read file: <nil>, using expression from flatfile: 30 4 * * *
configuration updated to file "/cron-schedule/..data". New cron expression: 30 4 * * *
no changes in schedule, nothing to do.
event not relevant: CREATE "/cron-schedule/config"
```

(`could not read file: <nil>` was also caused by shadowing the original `ReadFile` error with a successful `GetFlatIPAM` result; that logging issue is fixed in this PR.)

## Solution

- Add `ReconcilerConfigEventPredicate` that treats `CREATE` on both:
  - `..data` (repeat ConfigMap updates via atomic swap)
  - the projected config file path (first-time key projection)
- Use that helper from `cmd/reconciler`
- Fix the shadowed `err` in `determineCronExpression` so the real read failure is logged
- Add unit tests covering both relevant and irrelevant fsnotify events

No sleep/retry path: matching the first-time `config` CREATE is enough and avoids extra latency.

### Unit tests
```bash
go test ./pkg/reconciler/ -count=1
```
Passed locally.

### Manual reproduction on OpenShift (37-node cluster)

Cluster: **OCP 4.18.42**, OVN-Kubernetes, **37** `whereabouts-reconciler` pods (DaemonSet).

Steps:
1. Deployed Whereabouts via CNO `whereabouts-shim` additional network so the reconciler DaemonSet was running.
2. Created the schedule ConfigMap as documented:
   ```bash
   oc create configmap whereabouts-config -n openshift-multus \
     --from-literal=reconciler_cron_expression="* * * * *"
   ```
3. Checked all reconciler pods for whether they adopted `* * * * *`.

Result after settle:
| Outcome | Count |
|---|---|
| Updated successfully to `* * * * *` | **28 / 37** |
| Stuck on default `30 4 * * *` (bug) | **9 / 37** |

On affected pods, `/cron-schedule/config` already contained `* * * * *`, but the process still used `30 4 * * *` and logged the race signature above. Deleting/recreating an affected pod picked up the ConfigMap correctly (startup read path).

## Bug Fixes

* Improved configuration synchronization when projected configuration data or the configured file is created.
* Prevented unrelated file changes and removals from triggering unnecessary synchronization.
* Improved handling of configuration and scheduling errors to avoid applying invalid updates.
* Improved fallback configuration error reporting.

## Tests

* Added coverage for configuration event filtering and optional configuration mount scenarios.

### DIFF
--- a/pkg/reconciler/config_test.go
+++ b/pkg/reconciler/config_test.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 
 	"github.com/fsnotify/fsnotify"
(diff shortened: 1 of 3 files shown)


smulje:
Hi @dougbtv @maiqueb, could I please request your review of the PR when you get a chance? Thanks!

pkg/reconciler/config_test.go
```diff
@@ -72,6 +73,45 @@ var _ = Describe("Reconciler configuration watcher", func() {
...
+			Name: configPath,
+			Op:   fsnotify.Write,
+		}, false),
+		table.Entry("REMOVE on config file", fsnotify.Event{
+			Name: configPath,
+			Op:   fsnotify.Remove,
+		}, false),
+	)
+})
+
```
bpickard22, line 114: these tests are good adds, but the dont cover the regression described in the description. 
can you add a test the covers the kubelet emitting the create on `..data` before the symlink exists.

otherwise everything else looks good

smulje: Thanks for the review @bpickard22 
I added a regression suite in pkg/reconciler/config_test.go under ConfigMap first-projection race that exercises the same sequence described in the PR:

Start with the default schedule (30 4 * * *)
Simulate kubelet’s first ConfigMap projection: create the timestamped payload, swap ..data while the user-visible config symlink does not exist yet, wait in that race window, then create config → ..data/config with * * * * *
With ReconcilerConfigEventPredicate, assert the schedule updates to * * * * *
With the old ..data-only predicate, assert it stays stuck on 30 4 * * * (config CREATE ignored)
That matches the failure mode from the description: ..data CREATE is handled before config is readable, the read fails / falls back, and without treating config CREATE as relevant the cron never updates.

smulje:
@bpickard22 I have update test to accommodate the changes suggested. Please review.

PR #738, opened 2026-08-06
Proposal: first-available allocation across IP ranges

fabrizio2210:

Adds a draft proposal for an extensible IP-ranges allocation strategy. The proposed `ipRangesAllocation` field keeps the current one-address-per-range behavior for omission or `"all"`, and adds `"first_available"` to allocate one address total from the first configured range with capacity.

The proposal defines ordered fallback, typed exhaustion versus operational errors, idempotent retries across all pools, deterministic deallocation, legacy `range` ordering, overlapping ranges, and mixed-family behavior.

This extracts the single-address-across-ranges use case prototyped by @kruftik in #649 and credits that work in the proposal.

Related to #649; this PR does not close it.

This is a proposal-only change. Implementation and its unit/E2E coverage will follow after agreement on the API.

CIDR network/broadcast endpoint allocation and the related pool/IP-helper refactors from #649 are deliberately out of scope.

Validation: `git diff --cached --check`

## Summary by CodeRabbit

* **Documentation**
  * Added a draft proposal for an extensible IP range allocation strategy.
  * Documents configurable allocation modes, including first-available allocation.
  * Covers validation, range ordering, exhaustion handling, retries, deallocation, mixed address families, concurrency, compatibility, and range updates.
  * Includes implementation guidance, testing considerations, alternatives, and design decisions.
  * Clarifies how allocation behavior can remain backward compatible while supporting future strategies.

### DIFF
--- a/doc/proposals/ip-ranges-allocation-strategy.md
+++ b/doc/proposals/ip-ranges-allocation-strategy.md
@@ -1,0 +1,367 @@
+# IP ranges allocation strategy
+
+Status: Draft
+
+# Table of contents
+
+- [Introduction](#introduction)
+  - [Goals](#goals)
+  - [Non-goals](#non-goals)
+- [Design](#design)
+  - [Configuration](#configuration)
+  - [Range ordering](#range-ordering)
+  - [Allocation](#allocation)
+  - [Idempotent retries](#idempotent-retries)
+  - [Deallocation](#deallocation)
+  - [Overlapping ranges](#overlapping-ranges)
+  - [Mixed address families](#mixed-address-families)
+  - [Errors and concurrency](#errors-and-concurrency)
+  - [Backward compatibility](#backward-compatibility)
+  - [Changing strategies and ranges](#changing-strategies-and-ranges)
+- [Implementation plan](#implementation-plan)
+- [Test plan](#test-plan)
+- [Alternatives](#alternatives)
+- [Acknowledgements](#acknowledgements)
+- [Discussions and decisions](#discussions-and-decisions)
+
+<hr>
+
+## Introduction
+
+Whereabouts currently allocates one address from every entry in `ipRanges`.
+This is useful for dual-stack and multi-address attachments, but it cannot model
+one logical pool assembled from several ordered, disjoint ranges. In that use
+case, exhausting the first range should cause allocation to continue in the
+next range while the workload still receives exactly one address.
+
+This proposal adds an extensible allocation strategy to select between the
+current behavior and ordered, single-address fallback. It deliberately
+separates that API decision from the implementation.
+
+### Goals
+
+- Preserve the current one-address-per-range behavior by default.
+- Allow a configuration to allocate one address total from the first range
+  with capacity.
+- Define ordering, exhaustion, retries, deallocation, overlapping ranges, and
+  mixed-family behavior precisely.
+- Use an API that can be extended with additional allocation strategies later.
+
+### Non-goals
+
+- Allocating CIDR network or broadcast endpoint addresses.
+- Refactoring IP range helpers or introducing a new pool abstraction.
+- Changing IP selection within a range; Whereabouts continues to choose the
+  lowest available address according to the existing range configuration.
+- Changing the existing standard or Fast IPAM lease selection and datastore
+  concurrency models.
+- Implementing the feature in this proposal change. Implementation will follow
+  after maintainers agree on the API.
+
+<hr>
+
+## Design
+
+### Configuration
+
+Add the optional `ipRangesAllocation` string to the IPAM configuration. The
+initial supported values are:
+
+| Value | Behavior |
+| --- | --- |
+| omitted | Equivalent to `all`. |
+| `all` | Allocate one address from every configured range. This is the current behavior. |
+| `first_available` | Allocate exactly one address total from the first range with capacity. |
+
+Values are case-sensitive. An explicitly configured empty string or any value
+other than those listed above is a configuration error.
+
+For example, the following configuration treats two disjoint CIDRs as ordered
+capacity for a single attachment:
+
+```json
+{
+  "cniVersion": "0.3.1",
+  "name": "ordered-ranges",
+  "ipam": {
+    "type": "whereabouts",
+    "ipRanges": [
+      {
+        "range": "192.168.10.0/24"
+      },
+      {
+        "range": "192.168.20.0/24"
+      }
+    ],
+    "ipRangesAllocation": "first_available"
+  }
+}
+```
+
+The field describes how the `ipRanges` collection is consumed rather than a
+property of any individual range. A string strategy is preferred over a
+boolean such as `singleIP` so future policies can be added without introducing
+another set of interacting flags.
+
+### Range ordering
+
+The normalized `IPRanges` order is authoritative. Whereabouts already converts
+the legacy top-level `range` fields into a `RangeConfiguration` and prepends it
+to `IPRanges`. Therefore, when both forms are present, `first_available` tries
+the legacy `range` first and then the explicit `ipRanges` entries in their JSON
+order.
+
+Reordering ranges changes allocation preference but does not move existing
+allocations between pools.
+
+For `first_available`, Whereabouts resolves entries to their existing IPPool
+identifiers and visits each unique pool only once, at the position of its first
+configured occurrence. Preflight, fallback, duplicate refresh, and deallocation
+all use this de-duplicated traversal. Repeating an identical range with the same
+`network_name` therefore does not create an independent source of capacity.
+
+### Allocation
+
+The `all` strategy retains the existing allocation loop and its behavior.
+
+The `first_available` strategy performs these steps while holding the lease
+selected by the existing IPAM mode:
+
+1. Search every configured pool for an existing allocation for the same Pod
+   reference and interface, as described in [Idempotent retries](#idempotent-retries).
+2. If no existing allocation is found, visit ranges in normalized order.
+3. Attempt allocation from the current range using the existing within-range
+   address selection and overlapping-range checks.
+4. If the range reports typed exhaustion, continue to the next range.
+5. On the first successful allocation, persist it, return that one address,
+   and stop visiting ranges.
+6. If every range is exhausted, return an exhaustion error and no address.
+
+Fallback is only allowed for an error that specifically means the current
+range has no allocatable address. Parse, validation, API, authorization,
+timeout, conflict-after-retries, and other operational errors abort the
+request. Treating arbitrary failures as exhaustion could hide a broken pool
+and allocate from a lower-priority range unexpectedly.
+
+### Idempotent retries
+
+CNI ADD can be retried after the first attempt has persisted an allocation but
+before its result reaches the caller. A retry must return the original
+allocation instead of creating another one.
+
+Before attempting any new allocation, `first_available` searches the pool
+record for every currently configured range for a reservation whose `podRef`
+and interface name match the request. The search covers every configured range
+even if an earlier range currently has capacity. This is necessary for the
+following sequence:
+
+1. The first range is exhausted, so a Pod receives an address from the second
+   range.
+2. Capacity later becomes available in the first range.
+3. The Pod's CNI ADD is retried.
+
+Without a complete preflight search, step 3 would allocate a second address
+from the first range. With the search, Whereabouts returns the address already
+held in the second range and refreshes its container ID if needed, consistent
+with existing retry behavior within one pool.
+
+Failure to inspect any configured pool is an operational error; allocation
+must not proceed on incomplete knowledge. If inconsistent state contains more
+than one matching reservation, Whereabouts retryably refreshes the container
+ID on every match before returning the first match in configured order. If any
+refresh fails, ADD returns an operational error; a subsequent retry repeats the
+complete search and must not allocate another address. The inconsistency is
+reported in logs.
+
+When overlap protection is enabled, each unique preflight match must also pass
+through the existing overlapping-range reservation verification and update
+path before ADD succeeds. This repairs the failure window in which the IPPool
+allocation was persisted but its `OverlappingRangeIPReservation` was not. A
+missing reservation is recreated for the matching Pod and interface; a
+reservation owned by another workload is an inconsistency error. API or update
+failures are operational errors, and the existing IPPool allocation is not
+returned until overlap protection has been reconciled.
+
+### Deallocation
+
+Deallocation visits every configured pool in normalized order. Not finding the
+allocation in an earlier pool does not stop the search. Every reservation that
+matches the existing container-ID-and-interface deletion identity is removed,
+together with its overlapping-range reservation when enabled. Refreshing every
+duplicate match during ADD ensures a later DEL using the current container ID
+can remove all of them.
+
+Searching all pools makes deletion deterministic when the address came from a
+later fallback range. As today, deleting an allocation that is already absent
+succeeds.
+
+### Overlapping ranges
+
+The existing `enable_overlapping_ranges` behavior remains in force and remains
+scoped by `network_name`:
+
+- When overlap protection is enabled, an address reserved by another Pod is
+  unavailable even if it appears in another configured range. Allocation
+  continues within the current range, then falls back only if that range is
+  exhausted.
+- A retry that finds an existing IPPool allocation verifies or restores its
+  corresponding overlapping-range reservation before returning the address.
+- When overlap protection is disabled, `first_available` does not add a new
+  cross-range uniqueness guarantee; the current collision behavior is
+  preserved.
+
+No composite IPPool is introduced for overlapping entries. Their configured
+order remains the fallback order, and the existing range-to-IPPool mapping
+remains the unit of persistence. Identical entries that resolve to the same
+IPPool are visited once as described in [Range ordering](#range-ordering).
+
+### Mixed address families
+
+`first_available` means one address total, not one address per family. Given an
+IPv4 range followed by an IPv6 range, a request receives an IPv4 address while
+the first range has capacity and falls back to one IPv6 address only after the
+IPv4 range is exhausted.
+
+Dual-stack configurations that require both an IPv4 and an IPv6 address must
+omit `ipRangesAllocation` or set it to `all`.
+
+### Errors and concurrency
+
+Configuration parsing validates the strategy before any datastore operations.
+Exhaustion must be represented by a typed error so the range loop can
+distinguish expected capacity fallback from operational failure. If all ranges
+are exhausted, the returned error should identify total exhaustion and retain
+the per-range context needed for diagnostics.
+
+Standard IPAM continues to run the entire preflight search and
+allocation/deallocation operation under the current cluster-wide Whereabouts
+lease. Fast IPAM currently selects a node/pool-specific lease and does not yet
+support multiple ranges. Configuration validation therefore rejects combining
+`first_available` with `node_slice_size`. Supporting that combination requires
+a separate design for range ordering and cross-pool serialization.
+
+This proposal does not otherwise change lease selection. Existing datastore
+retry handling also remains in place; exhaustion is considered only after the
+current pool's transient retries are resolved.
+
+### Backward compatibility
+
+- Existing configurations omit `ipRangesAllocation` and therefore keep `all`.
+- Explicit `ipRangesAllocation: "all"` has the same behavior as omission.
+- When `node_slice_size` is not configured, configurations containing only the
+  legacy top-level `range` behave the same under either strategy because there
+  is only one normalized range.
+- When both `range` and `ipRanges` are configured, the legacy range remains
+  first, preserving the existing normalization order.
+- No IPPool or overlapping-range CRD schema changes or migrations are required.
+
+### Changing strategies and ranges
+
+Changing the strategy or reordering ranges affects new allocations only;
+Whereabouts does not migrate existing addresses between pools. If a workload
+already has matching reservations in more than one pool (for example, after
+changing from `all` to `first_available`), `first_available` returns the first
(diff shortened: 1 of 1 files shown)


doc/proposals/ip-ranges-allocation-strategy.md
```diff
@@ -0,0 +1,367 @@
...
+Deallocation visits every configured pool in normalized order. Not finding the
+allocation in an earlier pool does not stop the search. Every reservation that
+matches the existing container-ID-and-interface deletion identity is removed,
+together with its overlapping-range reservation when enabled. Refreshing every
+duplicate match during ADD ensures a later DEL using the current container ID
+can remove all of them.
+
+Searching all pools makes deletion deterministic when the address came from a
+later fallback range. As today, deleting an allocation that is already absent
+succeeds.
```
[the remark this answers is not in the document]
maiqueb: I agree with this comment.

doc/proposals/ip-ranges-allocation-strategy.md
```diff
@@ -0,0 +1,367 @@
...
+not create another reservation. A later CNI DEL searches all configured pools
+as described above.
+
+Removing or replacing a range changes the identity of the pool that allocation
+and deallocation inspect. Whereabouts cannot discover an allocation in a pool
+that is no longer represented in the configuration. Operators must therefore
+append replacement capacity while retaining old ranges until their IPPools
+have drained; only then may the old entries be removed. Removing or replacing
+a range that still has live allocations is unsupported and can otherwise
+produce a duplicate allocation on retry or an orphan on deletion.
```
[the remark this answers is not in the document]
maiqueb: what if we make the value immutable ? 

Would that workaround the issue ?

doc/proposals/ip-ranges-allocation-strategy.md
```diff
@@ -0,0 +1,367 @@
...
+initial supported values are:
+
+| Value | Behavior |
+| --- | --- |
+| omitted | Equivalent to `all`. |
+| `all` | Allocate one address from every configured range. This is the current behavior. |
+| `first_available` | Allocate exactly one address total from the first range with capacity. |
+
+Values are case-sensitive. An explicitly configured empty string or any value
+other than those listed above is a configuration error.
```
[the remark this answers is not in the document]
maiqueb: is there any value in disambiguating user specified "all" from default "all" ?

doc/proposals/ip-ranges-allocation-strategy.md
```diff
@@ -0,0 +1,367 @@
...
+    ],
+    "ipRangesAllocation": "first_available"
+  }
+}
+```
+
+The field describes how the `ipRanges` collection is consumed rather than a
+property of any individual range. A string strategy is preferred over a
+boolean such as `singleIP` so future policies can be added without introducing
+another set of interacting flags.
```
maiqueb, line 142: 100% . Thanks for taking that into account.

doc/proposals/ip-ranges-allocation-strategy.md
```diff
@@ -0,0 +1,367 @@
...
+### Range ordering
+
+The normalized `IPRanges` order is authoritative. Whereabouts already converts
+the legacy top-level `range` fields into a `RangeConfiguration` and prepends it
+to `IPRanges`. Therefore, when both forms are present, `first_available` tries
+the legacy `range` first and then the explicit `ipRanges` entries in their JSON
+order.
+
+Reordering ranges changes allocation preference but does not move existing
+allocations between pools.
```
maiqueb: the more I think about this issue, the more I believe we should have the range configuration immutable. Everything becomes simpler.

With that in mind, I believe we should add a `Use cases` section - so we know if the use cases we want to tackle are all covered.

doc/proposals/ip-ranges-allocation-strategy.md
```diff
@@ -0,0 +1,367 @@
...
+
+For `first_available`, Whereabouts resolves entries to their existing IPPool
+identifiers and visits each unique pool only once, at the position of its first
+configured occurrence. Preflight, fallback, duplicate refresh, and deallocation
+all use this de-duplicated traversal. Repeating an identical range with the same
+`network_name` therefore does not create an independent source of capacity.
```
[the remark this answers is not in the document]
maiqueb: I think the bot's right. The actual ranges must be taken into account.

doc/proposals/ip-ranges-allocation-strategy.md
```diff
@@ -0,0 +1,367 @@
...
+Before attempting any new allocation, `first_available` searches the pool
+record for every currently configured range for a reservation whose `podRef`
+and interface name match the request. The search covers every configured range
+even if an earlier range currently has capacity. This is necessary for the
+following sequence:
+
+1. The first range is exhausted, so a Pod receives an address from the second
+   range.
+2. Capacity later becomes available in the first range.
+3. The Pod's CNI ADD is retried.
```
maiqueb: what do you actually mean here ? I mean, how can a pod's CNI ADD be re-tried (after it succeeded) ? I assume it succeeded because you mention "a pod receives an address from the second range".

I don't think you can have that according to the [CNI spec](https://www.cni.dev/docs/spec/#add-add-container-to-network-or-apply-modifications):
> A runtime should not call ADD twice (without an intervening DEL) for the same (CNI_CONTAINERID, CNI_IFNAME) tuple.

Are we speaking about a stateful set pod ?..

doc/proposals/ip-ranges-allocation-strategy.md
```diff
@@ -0,0 +1,367 @@
...
+refresh fails, ADD returns an operational error; a subsequent retry repeats the
+complete search and must not allocate another address. The inconsistency is
+reported in logs.
+
+When overlap protection is enabled, each unique preflight match must also pass
+through the existing overlapping-range reservation verification and update
+path before ADD succeeds. This repairs the failure window in which the IPPool
+allocation was persisted but its `OverlappingRangeIPReservation` was not. A
+missing reservation is recreated for the matching Pod and interface; a
+reservation owned by another workload is an inconsistency error. API or update
```
maiqueb: this will be tricky to implement. We need to be able to effectively rebuilt the actually state from the current runtime, in order to reconcile the actual state with the desired cluster state.

doc/proposals/ip-ranges-allocation-strategy.md
```diff
@@ -0,0 +1,367 @@
...
+
+### Mixed address families
+
+`first_available` means one address total, not one address per family. Given an
+IPv4 range followed by an IPv6 range, a request receives an IPv4 address while
+the first range has capacity and falls back to one IPv6 address only after the
+IPv4 range is exhausted.
+
+Dual-stack configurations that require both an IPv4 and an IPv6 address must
+omit `ipRangesAllocation` or set it to `all`.
```
maiqueb, line 294: thank you for thinking this thorough. It does make sense (to me at least). 

If we ever want to have that particular behavior, we can just add another allocation strategy (e.g. `first_available_prefer_dualstack` or something).

doc/proposals/ip-ranges-allocation-strategy.md
```diff
@@ -0,0 +1,367 @@
...
+
+### Changing strategies and ranges
+
+Changing the strategy or reordering ranges affects new allocations only;
+Whereabouts does not migrate existing addresses between pools. If a workload
+already has matching reservations in more than one pool (for example, after
+changing from `all` to `first_available`), `first_available` returns the first
+match in configured order after refreshing every match's container ID and does
+not create another reservation. A later CNI DEL searches all configured pools
+as described above.
```
maiqueb: are there use-cases for updating the allocation strategies / configured ranges ? 

I believe it would be a lot simpler if we could treat the configuration as immutable. 
(not sure how, since we don't actually have a CRD to hold the config).

doc/proposals/ip-ranges-allocation-strategy.md
maiqueb: one of the reasons I think we should strive for immutability. 

this is very easy to get wrong, and it'll take massive effort to understand what's wrong, re-write the confs again to something that will allow the roll-back. 

... sure, in a gitops world, it's just reverting a commit, but manually ... 

I just feel the flow is very error prone. I'd like us to first work on the use cases, to understand if can have range immutability.

doc/proposals/ip-ranges-allocation-strategy.md
```diff
@@ -0,0 +1,367 @@
...
+  configured occurrence;
+- mixed IPv4/IPv6 ranges producing one address total with `first_available`;
+- legacy `range` alone and combined with `ipRanges`;
+- appending and reordering ranges while an existing allocation remains
+  discoverable, followed by removal only after the old pool drains;
+- rejection of `first_available` combined with `node_slice_size`.
+
+The implementation change will run `make test`. The documented kind-based E2E
+suite will also run when locally feasible; otherwise the implementation PR will
+report that it relies on the upstream E2E workflow.
```
maiqueb: TBH I don't follow this sentence. 

I think you're expected to provide *both* unit tests, and e2e tests.
We can waive the e2es for good reasons though.

maiqueb (changes requested):
Initial review.

I believe we need to discuss (... again ...) the list of use cases - if only to know if we can assume immutability of the configuration - which would simplify the development stage.

doc/proposals/ip-ranges-allocation-strategy.md
kruftik: unfortunately, they only should not, and not "must not", we have seen such a behaviour (multiple ADD calls) if kubelet or CRI restarted, there were additional calls

doc/proposals/ip-ranges-allocation-strategy.md
```diff
@@ -0,0 +1,367 @@
...
+### Non-goals
+
+- Allocating CIDR network or broadcast endpoint addresses.
+- Refactoring IP range helpers or introducing a new pool abstraction.
+- Changing IP selection within a range; Whereabouts continues to choose the
+  lowest available address according to the existing range configuration.
+- Changing the existing standard or Fast IPAM lease selection and datastore
+  concurrency models.
+- Implementing the feature in this proposal change. Implementation will follow
+  after maintainers agree on the API.
```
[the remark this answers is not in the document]
fabrizio2210: Good catch — updated in f42d569b to make the proposal-only scope explicit:

> This proposal defines the allocation-strategy API only. Implementation will follow after maintainers agree on it.

doc/proposals/ip-ranges-allocation-strategy.md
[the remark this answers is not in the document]
fabrizio2210: Agreed. Updated in 4fa9487b. We do not retain a distinction between an omitted value and explicit `all` after parsing; both normalize to `all`.

Presence is retained only through validation and merging so an explicit empty string or unsupported value is rejected rather than treated as omission or replaced by a flat-file default. The proposal now specifies that each source is validated before merging, a valid inline value wins, the flat-file value is used only when inline omits the field, and omission from both defaults to `all`.

doc/proposals/ip-ranges-allocation-strategy.md
fabrizio2210: You're welcome!

doc/proposals/ip-ranges-allocation-strategy.md
fabrizio2210: Thanks — I re-read the motivating discussion in [PR #649](https://github.com/k8snetworkplumbingwg/whereabouts/pull/649#issuecomment-3748492842) and checked analogous pool-IPAM behavior. I agree that the proposal should not support live strategy changes, reordering, editing, replacement, or removal.

The one live range update required by that use case is appending a newly assigned disjoint block after the original block is exhausted. I updated the proposal to define this contract:

- `ipRangesAllocation` and the existing normalized range sequence are immutable while allocations exist.
- `first_available` may append ranges at the end; the existing entries and their order remain immutable.
- `all` must be drained before appending too, because a retried ADD would otherwise allocate an additional address from the new range.
- Every other strategy or range-sequence change requires a full drain.
- Since this configuration lives in CNI configuration, commonly in a NetworkAttachmentDefinition, rather than a dedicated CRD, this is an operational invariant unless configuration generations are persisted.

I also added the requested Use cases section and updated the range-ordering, implementation-plan, and test-plan wording. Addressed in 6c4811db.

doc/proposals/ip-ranges-allocation-strategy.md
fabrizio2210: Thanks — the scenario is not specific to StatefulSets, and I agree that the previous example could be read as a normally acknowledged ADD being repeated for the same CNI tuple.

The intended case is recovery after an uncertain result or a kubelet/CRI restart. An earlier ADD may have persisted the allocation while its result did not reach the runtime, leaving the outcome uncertain. Alternatively, after a restart the runtime may create a replacement sandbox and invoke ADD again for the same Pod and interface without an intervening DEL. In the latter case, CNI_CONTAINERID can differ, so this is not necessarily a repeated ADD for the exact (CNI_CONTAINERID, CNI_IFNAME) tuple covered by the specification guidance. We have observed this kind of duplicate ADD behavior in practice, consistent with @kruftik’s point.

The idempotency check therefore searches every configured pool by podRef and interface, returns the existing allocation, and refreshes the container ID when necessary. I reworded the proposal and example to make these distinctions explicit in 8472698a.

doc/proposals/ip-ranges-allocation-strategy.md
fabrizio2210: Good point — the word “reconciled” made this sound broader than intended.

The ADD path should not attempt to rebuild runtime state or decide between conflicting owners. The intended repair is limited to completing the existing two-record allocation update. The proposal now requires a read-only validation pass across every matching IPPool allocation and its corresponding `OverlappingRangeIPReservation` before performing any repair or container-ID refresh.

If any overlap reservation belongs to a different Pod or interface, Whereabouts leaves the conflicting records unchanged and returns an operational inconsistency error. Only when validation finds no ownership conflict may an absent overlap reservation be recreated from the matching IPPool record and current ADD request. General reconstruction from Pods, network-status annotations, or the container runtime belongs in a broader reconciler and is outside this proposal.

I reworded the retry and overlapping-range sections, aligned the implementation plan, and added coverage for both missing-secondary repair and validation-before-mutation in 1dd02e44.

doc/proposals/ip-ranges-allocation-strategy.md
fabrizio2210: Agreed. Keeping first_available defined as exactly one address total gives it clear, deterministic semantics, while the extensible strategy field leaves room for a separate dual-stack-oriented strategy if that use case arises.

doc/proposals/ip-ranges-allocation-strategy.md
```diff
@@ -0,0 +1,431 @@
...
+The normalized allocation strategy and existing normalized range sequence are
+immutable while the network has live allocations. This means operators must
+not change between `all` and `first_available`, reorder ranges, edit an existing
+range, or remove or replace one. Such changes can alter an idempotent ADD
+result, make an existing reservation undiscoverable, or orphan it during DEL.
+
+For `first_available`, the only supported live update is appending one or more
+ranges. Existing entries and their order form an immutable prefix of the
+updated configuration. Appended ranges extend the pool for future allocations;
+they do not move existing allocations. The complete preflight search still
```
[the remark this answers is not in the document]
fabrizio2210: Good catch. Updated in 1e63bf94. Under first_available, an append increases capacity only for the first appended entry that resolves to a previously unseen IPPool identifier. An exact duplicate append remains a no-op and does not add capacity or alter the de-duplicated traversal. I clarified the range-ordering and live-update contracts and added test-plan coverage for both unique-pool expansion and duplicate append behavior. The complete consistency review also made network_name and other pool-identity inputs explicitly immutable while live allocations exist.

doc/proposals/ip-ranges-allocation-strategy.md
```diff
@@ -0,0 +1,367 @@
...
+Dual-stack configurations that require both an IPv4 and an IPv6 address must
+omit `ipRangesAllocation` or set it to `all`.
+
+### Errors and concurrency
+
+Configuration parsing validates the strategy before any datastore operations.
+Exhaustion must be represented by a typed error so the range loop can
+distinguish expected capacity fallback from operational failure. If all ranges
+are exhausted, the returned error should identify total exhaustion and retain
+the per-range context needed for diagnostics.
```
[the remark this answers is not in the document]
fabrizio2210: Good catch. Updated in b33d3c15. The existing `allocate.AssignmentError` is now the sole per-range fallback signal and is detected with `errors.As`; parse, validation, overlap, API, datastore, and other operational errors propagate unchanged and abort fallback. Total exhaustion returns a new `AllRangesExhaustedError` with ordered `RangeExhaustion` entries containing the first normalized configured index, resolved `PoolIdentifier`, and underlying `AssignmentError`; `Unwrap() []error` exposes those errors in traversal order. I also added coverage for direct and wrapped exhaustion, unchanged operational-error identity with no later-pool visit, and the ordered aggregate fields and unwrapping behavior.

doc/proposals/ip-ranges-allocation-strategy.md
fabrizio2210: Agreed. I checked the current repository setup, and the previous sentence conflated providing E2E coverage with running it locally. `make test` explicitly excludes the `e2e` packages, while GitHub Actions runs separate standard `e2e test` and `e2e test (node_slice)` KinD jobs for implementation changes. Updated in 62facac9: the implementation PR now explicitly requires both unit and E2E coverage, with core `first_available` behavior covered in the standard suite. Because `first_available` rejects `node_slice_size`, no feature-specific node-slice case is required, while the existing node-slice job remains regression coverage. Local KinD execution is conditional only on environment availability; the upstream standard E2E job is still expected to pass. Any waiver of new E2E coverage must be justified in the implementation PR and agreed by maintainers.

fabrizio2210:
Hi @maiqueb, I have updated the proposal to address the feedback from your initial review, especially the use cases and configuration immutability contract, along with the retry, error-handling, and test expectations. When you have time, could you please take another look? I left the relevant threads open so we can continue the discussion if anything still needs adjustment. Thanks!

doc/proposals/ip-ranges-allocation-strategy.md
```diff
@@ -0,0 +1,503 @@
...
+
+CNI ADD processing must be idempotent when the runtime invokes ADD again for
+the same Pod and interface. This can happen when an earlier ADD persisted an
+allocation but its result did not reach the runtime, leaving the outcome
+uncertain. It can also happen after a kubelet or CRI restart if the runtime
+creates a replacement sandbox and invokes ADD for the same Pod and interface
+without an intervening DEL. In the latter case, the `CNI_CONTAINERID` can
+differ, so this is not necessarily a repeated ADD for the exact
+`(CNI_CONTAINERID, CNI_IFNAME)` tuple that the CNI specification says a runtime
+should not issue. Neither case is specific to StatefulSets.
```
maiqueb: I am worried about these changes - from my perspective, this goes against the CNI spec:

1. [SPEC.md:247](https://github.com/containernetworking/cni/blob/3f51e8803ebbdba0ebeed735b42137e4c7302403/SPEC.md?plain=1#L247) => "If an interface of the requested name already exists in the container, the CNI plugin MUST return with an error."
2. [SPEC.md:249](https://github.com/containernetworking/cni/blob/3f51e8803ebbdba0ebeed735b42137e4c7302403/SPEC.md?plain=1#L249) — "A runtime should not call ADD twice (without an intervening DEL) for the same (CNI_CONTAINERID, CNI_IFNAME) tuple."

IMHO, the scenario you describe in
> This can happen when an earlier ADD persisted an allocation but its result did not reach the runtime, leaving the outcome uncertain.

does not "count" since the "earliear ADD" never completed - it needs to return its result (error code + stdout) to the runtime.

> It can also happen after a kubelet or CRI restart if the runtime
creates a replacement sandbox and invokes ADD for the same Pod and interface
without an intervening DEL. In the latter case, the `CNI_CONTAINERID` can
differ, so this is not necessarily a repeated ADD for the exact
`(CNI_CONTAINERID, CNI_IFNAME)` tuple that the CNI specification says a runtime
should not issue.

And this one - as you mention - is setting up networking on a separate sandbox, so, I also don't see the problem.

I just think we need to slight adapt the text to make it compliant with the SPEC.

fabrizio2210: I will reword the section around that replacement-sandbox case, remove the uncertain-result and same-tuple retry claims, and explicitly state that the upper plugin remains responsible for the CNI ADD requirements.

doc/proposals/ip-ranges-allocation-strategy.md
```diff
@@ -183,20 +184,24 @@ timeout, conflict-after-retries, and other operational errors abort the
...
+`(CNI_CONTAINERID, CNI_IFNAME)` tuple.
+
+Whereabouts already uses the Pod reference and interface name to find an
+existing allocation within one pool. When the upper CNI plugin configures a
+replacement sandbox for the same logical attachment, `CNI_CONTAINERID` and
+`CNI_NETNS` differ while the Pod reference and `CNI_IFNAME` remain the same.
+The requested interface does not already exist in the new sandbox. In this
+case, delegated Whereabouts ADD preserves the existing behavior: it returns the
+existing allocation and refreshes its stored container ID instead of allocating
+another address. This case is not specific to StatefulSets.
```
maiqueb, line 207: Nice, I like this description way more.

Also thank you for "isolating" the stateful sets scenario. 

Do you think it makes sense to capture that in the "non-goals" section ? Or, mention it as a future goal (which would require a separate enhancement) ?

fabrizio2210: Yes, I will add something in the non-goal section

PR #748, opened 2026-09-03
Fix pool.Update error fallthrough to ORIP update

### LINKED ISSUE #712: [BUG] IPPool allocation lost during concurrent pod creation, index missing from IPPool.spec.allocations despite OverlappingRangeIPReservation existing
**Describe the bug**
When multiple whereabouts pods allocate IPs concurrently from the same pool, a race condition can cause an allocation to be recorded in the `OverlappingRangeIPReservation` CR but **not** in the `IPPool.spec.allocations`. This leads to orphaned `OverlappingRangeIPReservation` entries that consume IP addresses but are not tracked in the IPPool, causing IP address leakage and eventual pool exhaustion.
Error that we see:
```
"failed to update the reservation list: the server rejected our request due
  to an error in our request"
```

**Expected behavior**
When pool.Update() fails after exhausting retries or encountering a non-temporary error:

- The OverlappingRangeIPReservation should NOT be created
- The function should return the error to the caller
- CNI should retry the entire allocation from scratch

The counts of IPPool.spec.allocations and OverlappingRangeIPReservation CRs should always be consistent.

**To Reproduce**
1. Create multiple IPPool ranges (3-4 pools, mix of IPv4 /24 and IPv6 /116 subnets)
2. Deploy 10+ pods using a Deployment with network-attachment-definitions referencing all pools
3. Wait for all pods to reach Running state
4. Compare IPPool.spec.allocations entries against OverlappingRangeIPReservation CRs

**Environment**:
- Whereabouts version : 0.9.3
- Kubernetes version (use `kubectl version`): N/A
- Others: N/A

**Additional info / context**
Add any other information / context about the problem here.

Ultimate-etamitlU:
## Summary

- Gate ORIP update on successful `pool.Update()` — add `err != nil`
  and `ctx.Err() != nil` checks after RETRYLOOP exit in
  `IPManagementKubernetesUpdate()`
- Return immediately on context cancellation inside the retry loop,
  before any pool mutation occurs
- Add three regression tests covering all failure modes

Fixes #712

## Problem

After the RETRYLOOP in `IPManagementKubernetesUpdate()`, there is no
error check before the `OverlappingRanges` block. When `pool.Update()`
fails (permanent error or temporary errors exhausting all retries),
execution falls through and creates an ORIP without a corresponding
IPPool allocation entry.

Additionally, context cancellation inside the retry loop used
`break RETRYLOOP`, which fell through to the ORIP update block
instead of returning immediately.

This causes IPPool/ORIP drift:
- The stale ORIP blocks future legitimate allocation of the same IP
  from overlapping ranges
- If the reconciler later cleans one CR but not the other, the
  inconsistency persists

## Fix

Three changes in `IPManagementKubernetesUpdate()`:

1. Inside the RETRYLOOP, context cancellation now returns immediately
   instead of breaking out of the loop, ensuring no pool mutation or
   ORIP creation occurs:

```go
select {
case <-ctx.Done():
    return newips, ctx.Err()
default:
}
```

2. After the RETRYLOOP, two guard clauses prevent ORIP creation when
   `pool.Update()` has failed or the context was canceled after the
   pool update completed:

```go
if err != nil {
    return newips, err
}
if ctx.Err() != nil {
    return newips, ctx.Err()
}
```

## Related work

The reconciler's orphan detection logic has a separate gap during
StatefulSet pod replacement windows that can independently cause
allocation inconsistency. That will be addressed in a follow-up issue.

## Test plan

- [x] `TestIPManagementKubernetesUpdate_PoolUpdateFailurePreventsORIPUpdate`
  — permanent `pool.Update()` error returns error, zero ORIP calls
- [x] `TestIPManagementKubernetesUpdate_RetriesExhaustedPreventsORIPUpdate`
  — temporary errors exhaust all retries, returns error, zero ORIP calls
- [x] `TestIPManagementKubernetesUpdate_ContextCanceledPreventsORIPUpdate`
  — context canceled during successful `pool.Update()`, returns
  `context.Canceled`, one IPPool patch completed, zero ORIP calls
- [x] Full `pkg/storage/kubernetes/` test suite passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of canceled operations during IP pool updates.
  - Prevented overlapping IP range reservations when pool updates fail or are canceled.
  - Ensured update errors are returned before any subsequent allocation changes are processed.

- **Tests**
  - Added coverage for permanent update failures, exhausted retries, and context cancellation.

### DIFF
--- a/pkg/storage/kubernetes/ipam.go
+++ b/pkg/storage/kubernetes/ipam.go
@@ -575,7 +575,7 @@ func IPManagementKubernetesUpdate(ctx context.Context, mode int, ipam *Kubernete
 		for j := 0; j < storage.DatastoreRetries; j++ {
 			select {
 			case <-ctx.Done():
-				break RETRYLOOP
+				return newips, ctx.Err()
 			default:
 				// retry the IPAM loop if the context has not been cancelled
 			}
@@ -694,6 +694,13 @@ func IPManagementKubernetesUpdate(ctx context.Context, mode int, ipam *Kubernete
 				break RETRYLOOP
 			}
 			break RETRYLOOP
+		}
+
+		if err != nil {
+			return newips, err
+		}
+		if ctx.Err() != nil {
+			return newips, ctx.Err()
 		}
 
 		if ipamConf.OverlappingRanges {
--- a/pkg/storage/kubernetes/ipam_test.go
+++ b/pkg/storage/kubernetes/ipam_test.go
@@ -1,7 +1,23 @@
 package kubernetes
 
-import "testing"
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sync/atomic"
+	"testing"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	k8stesting "k8s.io/client-go/testing"
+
+	whereaboutsv1alpha1 "github.com/k8snetworkplumbingwg/whereabouts/pkg/api/whereabouts.cni.cncf.io/v1alpha1"
+	fakewbclient "github.com/k8snetworkplumbingwg/whereabouts/pkg/generated/clientset/versioned/fake"
+	"github.com/k8snetworkplumbingwg/whereabouts/pkg/storage"
+	whereaboutstypes "github.com/k8snetworkplumbingwg/whereabouts/pkg/types"
+)
+
+// TestIPPoolName verifies IPPool CR naming for various network name and node name combinations.
 func TestIPPoolName(t *testing.T) {
 	cases := []struct {
 		name           string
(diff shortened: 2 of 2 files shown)


Ultimate-etamitlU:
> In `@pkg/storage/kubernetes/ipam_test.go`:
> - Around line 214-215: Update the cancellation test around ipam.Status to cancel
> the context from the successful “patch” reactor and let the fake tracker
> complete the patch, rather than pre-canceling before the status call. Ensure the
> test reaches the post-loop ctx.Err() guard and fails if that guard is removed.

Fixed with recent commit c509f89 !

Ultimate-etamitlU:
@coderabbit review --agent

Ultimate-etamitlU:
@coderabbitai review

